PR #2, opened 2021-10-13
Update geoparquet spec

TomAugspurger:
First pass at a spec for geoparquet format. Lots of open questions, but the basic idea is to encode geometry columns as WKB, and set a standard place for some additional metadata (like the CRS).

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -7,7 +7,7 @@ should be stored in parquet format, including the representation of geometries a
 
 ## Geometry columns
 
-Geometry columns are stored using the `BYTE_ARRAY` parquet type. They are encoded as WKB.
+Geometry columns are stored using the `BYTE_ARRAY` parquet type. They are encoded as [WKB](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary).
 
 ## Metadata
 
@@ -18,11 +18,18 @@ geoparquet files include additional metadata at two levels:
 
 ## File metadata
 
+All file-level metadata should be included under the "geoparquet" key in the parquet metadata.
+
 |     Field Name     |  Type  |                             Description                              |
 | ------------------ | ------ | -------------------------------------------------------------------- |
 | geoparquet_version | string | **REQUIRED** The version of the metadata standard used when writing. |
-| primary_column     | string | The "primary" geometry column.                                       |
+| primary_column     | string | The name of the "primary" geometry column.                           |
 
+
+### Additional file metadata information
+
+**primary_column**: This indicates the "primary" or "active" geometry for systems that can store multiple geometries,
+but have a default geometry used for geospatial operations.
 
 ## Column metadata
 
@@ -30,10 +37,7 @@ Each geometry column should include additional metadata
 
 | Field Name |                               Type                                      |                                                                   Description                                                                     |
 | ---------- | ----------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
-| epsg       | integer\| null                                                          | **REQUIRED.** [EPSG code](http://www.epsg-registry.org/) of the datasource                                                                        |
-| bbox       | \[number]                                                               | Bounding Box of the asset represented by this Item, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5).  |                                                                                                                                                                                      |
-| wkt2       | string\| null                                                           | [WKT2](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) string representing the Coordinate Reference System (CRS) used by the geometries |
-| projjson   | [PROJJSON Object](https://proj.org/specifications/projjson.html)\| null | PROJJSON object representing the Coordinate Reference System (CRS) that the `proj:geometry` and `proj:bbox` fields represent                      |
+| epsg       | integer\| null                                                          | **REQUIRED.** [EPSG code](http://www.epsg-registry.org/) of the datasource. Must be 4326                                                          |
 
 ### Additional information
 


format-specs/geoparquet.md
```diff
@@ -1,5 +1,47 @@
...
-## A Geospatial Parquet format
+## Overview
 
-This page is a scratch pad for ideas to define a 'geo-parquet' format, aligning on a default schema and format for geospatial data in Parquet.
+The [Apache Parquet][parquet] provides a standardized open-source columnar storage format. This specification defines how geospatial data
+should be stored in parquet format, including the representation of geometries and the required additional metadata.
 
+## Geometry columns
+
+Geometry columns are stored using the `BYTE_ARRAY` parquet type. They are encoded as WKB.
```
TomAugspurger, line 10: Is there an official specification for WKB to link to?

format-specs/geoparquet.md
```diff
@@ -1,5 +1,47 @@
...
+
+1. File metadata indicating things like the version of this specification used
+2. Column (chunk) metadata with additional metadata for each geometry column
+
+## File metadata
+
+|     Field Name     |  Type  |                             Description                              |
+| ------------------ | ------ | -------------------------------------------------------------------- |
+| geoparquet_version | string | **REQUIRED** The version of the metadata standard used when writing. |
+| primary_column     | string | The "primary" geometry column.                                       |
```
TomAugspurger, line 24: I'm not sure if this is appropriate to include or not.

Systems like geopandas (and maybe PostGIS?) can store multiple geometry dtype columns, but has the idea of an "active" or "primary" geometry that's used by default in geospatial operations.

If we include it, we'll need to update the description / add a section explaining it.

format-specs/geoparquet.md
```diff
@@ -1,5 +1,47 @@
...
+
+
+## Column metadata
+
+Each geometry column should include additional metadata
+
+| Field Name |                               Type                                      |                                                                   Description                                                                     |
+| ---------- | ----------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
+| epsg       | integer\| null                                                          | **REQUIRED.** [EPSG code](http://www.epsg-registry.org/) of the datasource                                                                        |
```
TomAugspurger, line 33: Mostly copied from https://github.com/stac-extensions/projection.

Do we want this here? Or do we keep the core spec lean and leave this to extensions?

format-specs/geoparquet.md
cholmes: I think it makes sense to have. I'm actually not sure of the precedence in different formats, but I believe most OGC api's/services allow multiple geometries, and also have operations that imply a 'default' geometry column. So definitely seems useful to me for the format to indicate what the default column is, especially to be used in a 'cloud-native' / data at-rest type of way.

format-specs/geoparquet.md
TomAugspurger: Sounds good.

> For the immediate term I think we just use epsg, and say it's required and must be set to 4326, and other options may be defined in extensions.

I vaguely recall discussions about non-geospatial geometries, i.e. no epsg / CRS. I don't recall if that was on the opening-call or another context, but should we allow `espg: null` to allow for this use case?

format-specs/geoparquet.md
```diff
@@ -1,5 +1,51 @@
...
+## Metadata
+
+geoparquet files include additional metadata at two levels:
+
+1. File metadata indicating things like the version of this specification used
+2. Column (chunk) metadata with additional metadata for each geometry column
+
+## File metadata
+
+All file-level metadata should be included under the "geoparquet" key in the parquet metadata.
```
TomAugspurger, line 21: This is new since your last review @cholmes. The motivation is to namespace things under "geoparquet" to avoid clashing with other systems using parquet metadata.

format-specs/geoparquet.md
```diff
@@ -1,5 +1,51 @@
...
+
+
+### Additional file metadata information
+
+**primary_column**: This indicates the "primary" or "active" geometry for systems that can store multiple geometries,
+but have a default geometry used for geospatial operations.
+
+## Column metadata
+
+Each geometry column should include additional metadata
```
TomAugspurger, line 36: I'm just noticing that https://github.com/geopandas/geo-arrow-spec/pull/2 puts the per-column metadata in the file metadata under `geo.columns`, which is a mapping from column name to additional metadata.

@brendan-ward do you know why that uses file metadata, rather than putting it in the column metadata for each column? I suppose it's a bit easier for tools to parse when it's in the file metadata, since they don't have to check the metadata for each column?

format-specs/geoparquet.md
```diff
@@ -1,5 +1,51 @@
...
+1. File metadata indicating things like the version of this specification used
+2. Column (chunk) metadata with additional metadata for each geometry column
+
+## File metadata
+
+All file-level metadata should be included under the "geoparquet" key in the parquet metadata.
+
+|     Field Name     |  Type  |                             Description                              |
+| ------------------ | ------ | -------------------------------------------------------------------- |
+| geoparquet_version | string | **REQUIRED** The version of the metadata standard used when writing. |
```
cholmes, line 25: Meant to ask earlier if you had a version for this, but looks like you did pick a 0.1.0 in your example. We should put some mention at the top about 'current version'. That it's an unreleased 0.1.0...

format-specs/geoparquet.md
jorisvandenbossche: Another question related to WKB is also which "flavour" of WKB is used (eg EWKB). I personally don't really know how important this is though (eg for GeoPandas, being backed by GEOS for WKB parsing, that automatically parses different flavors).

format-specs/geoparquet.md
jorisvandenbossche: Additional note on this (I only realized this last week): a good reason we used file-level metadata and not field-level metadata, is that Arrow doesn't actually support reading/writing field-level metadata for Parquet. 

The Parquet spec has this, so this could be added to the Arrow implementation. But even then, this might be less ideal. In the Parquet spec this lives in the ColumnMetaData of a ColumnChunk of a RowGroup (see https://parquet.apache.org/documentation/latest/#metadata). So that would mean that the metadata is repeated for each RowGroup, and a single Parquet file can consist of many row groups.

PR #4, opened 2021-10-19
Attempt to align with geoarrow spec

cholmes:
This PR aims to align the geoparquet with the [GeoArrow proposal](https://github.com/geopandas/geo-arrow-spec/pull/2). We hope to fully align, with this spec as the parquet instantiation of all the same geoarrow principles. 

For now I tried to align all the names / structure. @TomAugspurger - would be great if you could check if I got it right. They use a JSON description, which is nice, but wouldn't actually be used in parquet right? So I think we should describe it without that, though might be nice to have some sort of 'example' that is just the tables as they'd be in parquet?

I did slim a couple things they have:
* Removed library and version of the 'creator'. Mostly because I've not seen any precedence for this in geospatial software, so I think many implementors might be a bit confused, and may not keep it up to date, etc. But I do think it's a cool idea, so would be open to making it optional? Either way should give people guidance on what they put there.
* Made CRS just wkt2 - I think the spec should be narrow on this, as not everybody will use the proj library. And then strongly recommended that it's always 4326.

I also changed geoparquet_version to be just 'version'. They had a 'schema_version' in the spec, but then just 'version' in the example. I liked just 'version'. Though maybe we should have a specific geoparquet version? 

I also left out the bounding box stuff, though it does seem interesting. If we want to explain how it'd work for parquet partitioning use cases then I'd be for having it in.

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -28,9 +28,9 @@ All file-level metadata should be included under the "geo" key in the parquet me
 
 |     Field Name     |  Type  |                             Description                              |
 | ------------------ | ------ | -------------------------------------------------------------------- |
-| version     | string | **REQUIRED** The version of the geoparquet metadata standard used when writing.   |
+| version     		 | string | **REQUIRED** The version of the geoparquet metadata standard used when writing.   |
 | primary_column     | string | **REQUIRED** The name of the "primary" geometry column.                |
-| columns            | \[[Column Metadata](#column-metadata)\]  | **REQUIRED** Metadata about geometry columns. |
+| columns            | Map<key, [Colum Metadata](#column-metadata)> | **REQUIRED** Metadata about geometry columns, with each key is the name of a geometry column in the table. |
 
 
 ### Additional file metadata information
@@ -40,7 +40,7 @@ All file-level metadata should be included under the "geo" key in the parquet me
 This indicates the "primary" or "active" geometry for systems that can store multiple geometries,
 but have a default geometry used for geospatial operations.
 
-#### schema_version
+#### version
 
 Version of the geoparquet spec used, currently 0.1.0
 


format-specs/geoparquet.md
```diff
@@ -14,38 +18,73 @@ Geometry columns are stored using the `BYTE_ARRAY` parquet type. They are encode
...
 ### Additional file metadata information
 
-**primary_column**: This indicates the "primary" or "active" geometry for systems that can store multiple geometries,
+#### primary_column
+
+This indicates the "primary" or "active" geometry for systems that can store multiple geometries,
 but have a default geometry used for geospatial operations.
 
-## Column metadata
+#### schema_version
```
TomAugspurger, line 43: This should be `version`, to match the field name in the table?

format-specs/geoparquet.md
```diff
@@ -14,38 +18,73 @@ Geometry columns are stored using the `BYTE_ARRAY` parquet type. They are encode
...
-All file-level metadata should be included under the "geoparquet" key in the parquet metadata.
+All file-level metadata should be included under the "geo" key in the parquet metadata.
 
 |     Field Name     |  Type  |                             Description                              |
 | ------------------ | ------ | -------------------------------------------------------------------- |
-| geoparquet_version | string | **REQUIRED** The version of the metadata standard used when writing. |
-| primary_column     | string | The name of the "primary" geometry column.                           |
+| version     | string | **REQUIRED** The version of the geoparquet metadata standard used when writing.   |
+| primary_column     | string | **REQUIRED** The name of the "primary" geometry column.                |
+| columns            | \[[Column Metadata](#column-metadata)\]  | **REQUIRED** Metadata about geometry columns. |
```
TomAugspurger, line 33: I think this type should be something like `Map<key, [Colum Metadata](#column-metadata)>`, where the key is the name of the column in the table.

TomAugspurger (commented):
92be42b8626543ac48f10c69ec04576699462cd9 updated the example to match the new spec (with my other two comments).

I need to verify two things:

1. The WKT CRS is different

<details>

```
GEOGCRS["WGS 84",ENSEMBLE["World Geodetic System 1984 ensemble",MEMBER["World Geodetic System 1984 (Transit)"],MEMBER["World Geodetic System 1984 (G730)"],MEMBER["World Geodetic System 1984 (G873)"],MEMBER["World Geodetic System 1984 (G1150)"],MEMBER["World Geodetic System 1984 (G1674)"],MEMBER["World Geodetic System 1984 (G1762)"],ELLIPSOID["WGS 84",6378137,298.257223563,LENGTHUNIT["metre",1]],ENSEMBLEACCURACY[2.0]],PRIMEM["Greenwich",0,ANGLEUNIT["degree",0.0174532925199433]],CS[ellipsoidal,2],AXIS["geodetic latitude (Lat)",north,ORDER[1],ANGLEUNIT["degree",0.0174532925199433]],AXIS["geodetic longitude (Lon)",east,ORDER[2],ANGLEUNIT["degree",0.0174532925199433]],USAGE[SCOPE["Horizontal component of 3D system."],AREA["World."],BBOX[-90,-180,90,180]],ID["EPSG",4326]]
```

</details>

2. I need to confirm that this is actually written to the parquet metadata, and not just the pyarrow metadata (doesn't have to hold up this PR).

TomAugspurger:
On the bounding boxes, there's a bit of motivation at https://github.com/geopandas/geo-arrow-spec/pull/2/files#diff-6cf20cd11635a4f1fc656313a8d9e2a14f62dc7b68d778deac688b4dd051ffe2R52.

@jorisvandenbossche, do you know if dask-geopandas is actually using `bbox` from the draft geo-arrow-spec? It seems like it's not. IIUC it's using a separate top-level key: https://github.com/geopandas/dask-geopandas/blob/393dcb558dfaa8bfbfc0cd98a8a52891b8046186/dask_geopandas/io/parquet.py#L142 (I don't really know how file-level metadata like this interacts with the parquet dataset and separate partitions of that dataset).

format-specs/geoparquet.md
```diff
@@ -5,6 +5,10 @@
 The [Apache Parquet][parquet] provides a standardized open-source columnar storage format. This specification defines how geospatial data
 should be stored in parquet format, including the representation of geometries and the required additional metadata.
 
+## Version
+
+This is version 0.1.0 of the geoparquest specification.
```
jorisvandenbossche, line 10: geoparquest -> geoparquet

format-specs/geoparquet.md
```diff
@@ -14,38 +18,73 @@ Geometry columns are stored using the `BYTE_ARRAY` parquet type. They are encode
...
 
 | Field Name |                               Type                                      |                                                                   Description                                                                     |
 | ---------- | ----------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
-| epsg       | integer\| null                                                          | **REQUIRED.** [EPSG code](http://www.epsg-registry.org/) of the datasource. Must be 4326                                                          |
+| crs       | string   | **REQUIRED** [WKT2](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) string representing the Coordinate Reference System (CRS) of the geometry.  |
+| encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported. |
+
+#### crs
+
+It is strongly recommended to use [EPSG:4326 (lat, long)](https://spatialreference.org/ref/epsg/4326/) for all data, so in most cases the value of the crs should be:
```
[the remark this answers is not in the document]
cholmes: Yeah, I'm not 100% sure of the default CRS to use. If WKB is long, lat then I agree it makes sense to use crs84. @ghobona - what's the wkt2 for crs84? And is there an entry on spatialreference.org? It'd be nice to link to a web page and not just some big xml document.

PR #5, opened 2021-11-18
Align "geo" key in metadata and example

jorisvandenbossche:
See https://github.com/opengeospatial/cdw-geo/pull/4#discussion_r733968435

The metadata description was updated to use "geo", but the example this uses "geoparquet"


TomAugspurger:
Thanks!

Issue #14, opened 2022-03-01
Clarify WKB encoding type


TomAugspurger:
There are various types of WKB encoding. We should clarify exactly what flavor of WKB is expected (strict, extended, ...).

https://libgeos.org/specifications/wkb/

TomAugspurger:
According to https://libgeos.org/specifications/wkb/

> Standard WKB supports two-dimensional geometry, and is a proper subset of both extended WKB and ISO WKB.

Are we OK with that limitation?

cholmes:
Discussion on call is that we are ok with that limitation for 0.1, and then discuss in future iso and ewkb as options.

PR #16, opened 2022-03-01
Clarify that WKB means "standard" WKB enconding

TomAugspurger:
Closes https://github.com/opengeospatial/geoparquet/issues/14

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -76,7 +76,7 @@ Data that is better served in particular projections can choose to use an altern
 
 This is the binary format that the geometry is encoded in. The string 'WKB' to represent 
 [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
-of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/) WKB representation.
+of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
 
 ### Additional information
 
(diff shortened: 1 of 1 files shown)


format-specs/geoparquet.md
```diff
@@ -76,7 +76,7 @@ Data that is better served in particular projections can choose to use an altern
 
 This is the binary format that the geometry is encoded in. The string 'WKB' to represent 
 [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
-of the spec may support alternative encodings.
+of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/) WKB representation.
```
kylebarron, line 79: Link to the `#standard-wkb` section?
```suggestion
of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
```

Maybe I missed this on the call, but that means we're sticking to 2D coordinates for 0.1?

pomadchin (commented):
LGTM, (with the suggestion by @kylebarron). 
Also the commit message could be pretty confusing (i mean _Standard CRS for WKB_).

pomadchin (approved):

cholmes (approved):

jorisvandenbossche (approved):

jorisvandenbossche:
>  Also the commit message could be pretty confusing (i mean "Standard CRS for WKB").

Squash and merge should have solved that!

cholmes:
I noticed that squash and merge is the default and only enabled option now. Did someone change that? Or does github default to that now? Do we want squash & merge for all PR's? I'm happy for anything, just haven't done the squash and merge thing much before.

jorisvandenbossche:
I changed that (but happy to change it back). I personally think that's the easiest option (and gives the cleanest git history)

pomadchin:
I don't have strong feelings about it, but would prefer squash & merge as a default, it is easier to track the history of changes skipping extra commits.
Could be manually squashed if there are important commits to track in the history. ¯\_(ツ)_/¯

cholmes:
> I changed that (but happy to change it back). I personally think that's the easiest option (and gives the cleanest git history)

Cool. More than happy to keep it, seems cool, I've just never worked with it.

Issue #3, opened 2021-10-14
geoparquet coordinate reference system default / format / extension


cholmes:
There are a lot of options for how we approach coordinate reference systems.

* GeoJSON only allows 4326 data. They started with more options, but then narrowed it down.
* Simple Features for SQL defines an 'SRID table' where you are supposed to map number id's to crs well known text. PostGIS uses the same srid's as epsg, but oracle doesn't.
* WKT doesn't include projection info, but that's seen as a weakness, and one main reason why ewkt came. I believe ewkt just assumes epsg / postgis default srid table, so it's not fully interoperable with Oracle.
* STAC uses epsg, but you can set it to null and provide projjson or crs-wkt
* OGC API - Features core specifies just WGS-84 (long, lat), using a URI like http://www.opengis.net/def/crs/OGC/1.3/CRS84, see [crs info](https://github.com/cportele/ogcapi-building-blocks/blob/main/geo/common/json-collection.adoc#coordinate-reference-system-crs-identifiers)

And there's obviously more. 

My general take is that we should have a default, and expect most things to use that. But should specify it in a way that it could be an extension in the future. So we shouldn't just say 'everything is 4326' just in the spec, but should have a field that says this field is always 4326 for the core spec, but in the future that field could have other values.

So I think we do the first version with just 4326, and then when people ask for more we can have an extension.

One thing I'm not sure about is whether we should use 'epsg' as the field. EPSG covers most projections people want, but not all. In geopackage they just create a whole srid table to then refer to, so the SRID's used are defined. Usually the full epsg database is included, but then users can add other options. 

One potential option would be to follow ogc api - features and use URI's. I'm not sure how widely accepted that approach is, like if the full epsg database is already referenced online. So instead of 'epsg' as the field we'd have 'crs', and it's a string URI.

EFT-Defra:
Hi @cholmes, I'm a data engineer at the Department of Environment, Food, and Rural Affairs in the UK. We're currently in the process of moving our spatial data processing and analysis into the cloud, where we're hoping that Spark will help us to exploit data sources that were previously beyond the capabilities of our aging IT infrastructure. As such we've been watching the work that @jorisvandenbossche and the GeoPandas' team have been doing on geoparquet with anticipation, so we're really pleased that you're all working together on this. Is there anyway we can help or contribute?

cholmes:
[...]

PR #24, opened 2022-03-03
Update example parquet file

TomAugspurger:
* Adds `bbox`, consistent with https://github.com/opengeospatial/geoparquet/pull/21/
* Fixed the `pyarrow.parquet` import

xref https://github.com/opengeospatial/geoparquet/issues/22. Will leave that open till all the changes in 0.1 are final. I think the main outstanding issue is https://github.com/opengeospatial/geoparquet/issues/9. This file should still be compatible with that, but we'll want to add a second example with planar edges.

PR #25, opened 2022-03-03
CRS spec definition for version 0.1

alasarr:
This PR aims to cover the definition of CRS for version 0.1.

I've covered all the topics discussed at https://github.com/opengeospatial/geoparquet/issues/3


alasarr:
I've just made a change to the recommended CRS. I've decided to move from OGC:84 to EPSG:4326 (as we decided at https://github.com/opengeospatial/geoparquet/issues/3#issuecomment-1055799929). The reasons are:
- We decided OGC:84 over EPSG:4326 because of the [axis order](https://github.com/opengeospatial/geoparquet/issues/3#issuecomment-1045547954), but we later decided to force a specific axis order of long/lat.
- The first dataset I was about to use (at our [example.py](https://github.com/opengeospatial/geoparquet/blob/main/examples/geoparquet/example.py#L39) was in EPSG:4326 and I needed to transform it to OGC:84.  
- 4326 is far more used than OGC:84.
- The experience working with OGC:84 is harder than 4326. It's not very hard, but I played a little bit with geopandas and pyproj to get the WKT2 string, apply transformations, etc... and it was easier for me with 4326 because I'm more familiar with this, I think lots of users will have the same feeling.

Sorry to re-open this, but I think it's important. If you don't agree with this change we can rollback to OGC:84: https://github.com/opengeospatial/geoparquet/pull/25/commits/2ddcb907ee67dd80074a17835f872e1b2fe0aa37

cholmes:
I was actually looking around at OGC 84 as well and reached a similar conclusion, so I'm +1 on keeping EPSG:4326. And agree that our decision to force axis order also helps. I think we should perhaps 'cite' geopackage, refer to the section of the spec, and say that we are following them.

Perhaps it would be good to write up a small page in the spec repo that helps explain the state of things. It looks like geojson uses OGC:84, but it also looks like doing that is easier if you're not including a WKT2. I do worry a bit that we're just propagating the issue with using 4326 and doing lon/lat. Perhaps we should ask the geopackage group if they still feel good about their decision.

But yes, for 0.1 let's just go with 4326 and clarify that we're overriding it, and we can also revisit it.

format-specs/geoparquet.md
```diff
@@ -50,26 +50,43 @@ Each geometry column in the dataset must be included in the columns field above
...
-| crs       | string   | **REQUIRED** [WKT2](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) string representing the Coordinate Reference System (CRS) of the geometry.  |
+| crs       | string   | **REQUIRED**  string representing the Coordinate Reference System (CRS) of the geometry.  |
 | encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported. |
 
 #### crs
 
-It is strongly recommended to use [EPSG:4326 (lat, long)](https://spatialreference.org/ref/epsg/4326/) for all data, so in most cases the value of the crs should be:
+The Coordinate Reference System (CRS) is a mandatory parameter for all the geometries defined in geoparquet format. 
+
+The CRS needs to be provided in [WKT](https://en.wikipedia.org/wiki/Well-known_text_representation_of_coordinate_reference_systems) version 2, also known as **WKT2**. WKT2 has several revisions, this specification supports the revisions from [2015](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) and [2019](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html): WKT2_2015, WKT2_2015_SIMPLIFIED, WKT2_2019, WKT_2019_SIMPLIFIED.
```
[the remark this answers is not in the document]
cholmes: Added #26 for general discussion on officially adopting this kind of language.

format-specs/geoparquet.md
```diff
@@ -50,26 +50,43 @@ Each geometry column in the dataset must be included in the columns field above
...
 
 #### crs
 
-It is strongly recommended to use [EPSG:4326 (lat, long)](https://spatialreference.org/ref/epsg/4326/) for all data, so in most cases the value of the crs should be:
+The Coordinate Reference System (CRS) is a mandatory parameter for all the geometries defined in geoparquet format. 
+
+The CRS needs to be provided in [WKT](https://en.wikipedia.org/wiki/Well-known_text_representation_of_coordinate_reference_systems) version 2, also known as **WKT2**. WKT2 has several revisions, this specification supports the revisions from [2015](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) and [2019](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html): WKT2_2015, WKT2_2015_SIMPLIFIED, WKT2_2019, WKT_2019_SIMPLIFIED. 
+
+
+As the most common CRS for datasets is latitude/longitude, for the widest interoperability we recommend [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84) for all data, so in most cases the value of the crs should be:
```
cholmes, line 63: I feel like we should mention here that our axis order overrides this?

format-specs/geoparquet.md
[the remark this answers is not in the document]
cholmes: Do we need/want to support both 2015 and 2019? What are the advantages of doing so? In general I prefer just one way to do things, but am not an expert on these.

format-specs/geoparquet.md
```diff
@@ -50,26 +50,43 @@ Each geometry column in the dataset must be included in the columns field above
...
-    DATUM["WGS_1984",
-        SPHEROID["WGS 84",6378137,298.257223563,
-            AUTHORITY["EPSG","7030"]],
-        AUTHORITY["EPSG","6326"]],
-    PRIMEM["Greenwich",0,
-        AUTHORITY["EPSG","8901"]],
-    UNIT["degree",0.01745329251994328,
-        AUTHORITY["EPSG","9122"]],
-    AUTHORITY["EPSG","4326"]]
+GEOGCRS["WGS 84",
```
cholmes, line 66: Where did this definition come from? I got mine from https://spatialreference.org/ref/epsg/4326/ and they seem different?

format-specs/geoparquet.md
```diff
@@ -50,26 +50,43 @@ Each geometry column in the dataset must be included in the columns field above
...
+        UNIT["degree",0.0174532925199433],
+    USAGE[
+        SCOPE["Horizontal component of 3D system."],
+        AREA["World."],
+        BBOX[-90,-180,90,180]],
+    ID["EPSG",4326]]
 ```
 
+Due to the large number of CRSes available and the difficulty of implementing all of them, we strongly encourage the first implementing [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84).
+
```
cholmes, line 89: Tried to flesh this out a bit more.

```suggestion
Due to the large number of CRSes available and the difficulty of implementing all of them, we expect that a number of implementations will at least start with only support a single CRS. To maximize interoperability we strongly recommend GeoParquest tool providers to always implement support for [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84). 
```

format-specs/geoparquet.md
```diff
@@ -50,26 +50,43 @@ Each geometry column in the dataset must be included in the columns field above
...
 Data that is better served in particular projections can choose to use an alternate coordinate reference system.
```
cholmes, line 90:
```suggestion
Users are recommended to store their data in EPSG:4326 for it to work with the widest number of tools. But data that is better served in particular projections can choose to use an alternate coordinate reference system. We expect many tools will support alternate CRS's, but encourage users to check.
```

format-specs/geoparquet.md
```diff
@@ -78,6 +95,10 @@ This is the binary format that the geometry is encoded in. The string 'WKB' to r
 [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
 of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
 
+#### Coordinated order
```
cholmes, line 98:
```suggestion
#### Coordinate Order
```

format-specs/geoparquet.md
```diff
@@ -78,6 +95,10 @@ This is the binary format that the geometry is encoded in. The string 'WKB' to r
 [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
```
cholmes, line 95:
```suggestion
The string 'WKB', signifying [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
```

format-specs/geoparquet.md
```diff
@@ -78,6 +95,10 @@ This is the binary format that the geometry is encoded in. The string 'WKB' to r
...
 of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
```
cholmes, line 96:
```suggestion
of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB 
representation. This means 3D coordinates are not supported in this version of GeoParquet, but we expect
this to come in a future version.
```

cholmes (changes requested):
Overall looks good. Made a bunch of suggested commits and comments.

format-specs/geoparquet.md
pomadchin: @cholmes 

```bash
$ conda list | grep proj    
proj                      8.2.1                h1512c50_0    conda-forge
$ projinfo EPSG:4326            
PROJ.4 string:
+proj=longlat +datum=WGS84 +no_defs +type=crs

WKT2:2019 string:
GEOGCRS["WGS 84",
    ENSEMBLE["World Geodetic System 1984 ensemble",
        MEMBER["World Geodetic System 1984 (Transit)"],
        MEMBER["World Geodetic System 1984 (G730)"],
        MEMBER["World Geodetic System 1984 (G873)"],
        MEMBER["World Geodetic System 1984 (G1150)"],
        MEMBER["World Geodetic System 1984 (G1674)"],
        MEMBER["World Geodetic System 1984 (G1762)"],
        MEMBER["World Geodetic System 1984 (G2139)"],
        ELLIPSOID["WGS 84",6378137,298.257223563,
            LENGTHUNIT["metre",1]],
        ENSEMBLEACCURACY[2.0]],
    PRIMEM["Greenwich",0,
        ANGLEUNIT["degree",0.0174532925199433]],
    CS[ellipsoidal,2],
        AXIS["geodetic latitude (Lat)",north,
            ORDER[1],
            ANGLEUNIT["degree",0.0174532925199433]],
        AXIS["geodetic longitude (Lon)",east,
            ORDER[2],
            ANGLEUNIT["degree",0.0174532925199433]],
    USAGE[
        SCOPE["Horizontal component of 3D system."],
        AREA["World."],
        BBOX[-90,-180,90,180]],
    ID["EPSG",4326]]
    ```

Comment (alasarr) [reply format-specs/geoparquet.md:66]: It comes from pyproj:
```python
import pyproj
from pyproj.enums import WktVersion
from pyproj import CRS
crs = pyproj.CRS.from_epsg(4326)
print(crs.to_wkt(WktVersion.WKT2_2019_SIMPLIFIED, pretty=True))
```

format-specs/geoparquet.md
```diff
@@ -50,26 +50,43 @@ Each geometry column in the dataset must be included in the columns field above
 
 | Field Name |                               Type                                      |                                                                   Description                                                                     |
 | ---------- | ----------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
-| crs       | string   | **REQUIRED** [WKT2](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) string representing the Coordinate Reference System (CRS) of the geometry.  |
+| crs       | string   | **REQUIRED**  string representing the Coordinate Reference System (CRS) of the geometry.  |
```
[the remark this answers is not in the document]
alasarr: I wanted to update the link with 2019 revision and I completely remove it.

format-specs/geoparquet.md
alasarr: We already have a specific section for coordinate order where we specify that. Did you see it? Or you just want to emphasize here too.

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
+        UNIT["degree",0.0174532925199433],
+    USAGE[
+        SCOPE["Horizontal component of 3D system."],
+        AREA["World."],
+        BBOX[-90,-180,90,180]],
+    ID["EPSG",4326]]
 ```
 
-Data that is better served in particular projections can choose to use an alternate coordinate reference system.
+Due to the large number of CRSes available and the difficulty of implementing all of them, we expect that a number of implementations will at least start with only support a single CRS. To maximize interoperability we strongly recommend GeoParquest tool providers to always implement support for [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84).
```
jorisvandenbossche, line 88:
```suggestion
Due to the large number of CRSes available and the difficulty of implementing all of them, we expect that a number of implementations will at least start with only support a single CRS. To maximize interoperability we strongly recommend GeoParquet tool providers to always implement support for [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84). 
```

jorisvandenbossche: It's not really clear to me what is exactly meant with "tool providers" in this sentence.

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
 
 #### crs
 
-It is strongly recommended to use [EPSG:4326 (lat, long)](https://spatialreference.org/ref/epsg/4326/) for all data, so in most cases the value of the crs should be:
+The Coordinate Reference System (CRS) is a mandatory parameter for each geometry column defined in geoparquet format. 
+
+The CRS must be provided in [WKT](https://en.wikipedia.org/wiki/Well-known_text_representation_of_coordinate_reference_systems) version 2, also known as **WKT2**. WKT2 has several revisions, this specification supports the revisions from [2015](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) and [2019](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html): WKT2_2015, WKT2_2015_SIMPLIFIED, WKT2_2019, WKT_2019_SIMPLIFIED. 
+
+
+As the most common CRS for datasets is latitude/longitude, for the widest interoperability we recommend [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84) for all data, so in most cases the value of the crs should be:
```
jorisvandenbossche, line 63: We can maybe link to https://epsg.org/crs_4326/WGS-84.html instead of https://spatialreference.org (which is known to not be very up to date). 
Although that official page from the EPSG database doesn't include different representations (like WKT) as spatialreference.org or epsg.io do, but having no WKT instead of outdated WKT is maybe better ..

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
+    USAGE[
+        SCOPE["Horizontal component of 3D system."],
+        AREA["World."],
+        BBOX[-90,-180,90,180]],
+    ID["EPSG",4326]]
 ```
 
-Data that is better served in particular projections can choose to use an alternate coordinate reference system.
+Due to the large number of CRSes available and the difficulty of implementing all of them, we expect that a number of implementations will at least start with only support a single CRS. To maximize interoperability we strongly recommend GeoParquest tool providers to always implement support for [EPSG:4326](https://spatialreference.org/ref/epsg/wgs-84). 
+Users are recommended to store their data in EPSG:4326 for it to work with the widest number of tools. But data that is better served in particular projections can choose to use an alternate coordinate reference system. We expect many tools will support alternate CRS's, but encourage users to check.
```
jorisvandenbossche, line 89:
```suggestion
Users are recommended to store their data in EPSG:4326 for it to work with the widest number of tools. But data that is better served in particular projections can choose to use an alternate coordinate reference system. We expect many tools will support alternate CRSes, but encourage users to check.
```

Small language nit: we should probably choose between `CRS's` and `CRSes` as the plural form and use one consistently, as now we used both in the text here (not directly an idea which of the two is most correct)

EDIT, from a a quick google, it seems that CRSes is the correct way?

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
 
 #### encoding
 
 This is the binary format that the geometry is encoded in. The string 'WKB' to represent
```
jorisvandenbossche, line 93:
```suggestion
This is the binary format that the geometry is encoded in.
```

(after the previous edit, the start of this sentence is now duplicated)

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
-[Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
-of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
+The string 'WKB', signifying [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
+of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB 
+representation. This means 3D coordinates are not supported in this version of GeoParquet, but we expect
+this to come in a future version.
+
+#### Coordinate Order
+
+The axis order in WKB stored in a geoparquet follows the de facto standard for axis order in WKB and is therefore always (x,y{,z}{,m}) where x is easting or longitude, y is northing or latitude, z is optional elevation, and m is optional measure. This ordering explicitly overrides the axis order as specified in the CRS.
```
jorisvandenbossche, line 101:
```suggestion
The axis order of the coordinates in WKB stored in a geoparquet follows the de facto standard for axis order in WKB and is therefore always (x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS.
```

Since we don't yet support z or m values, we should probably leave that part out for now, to strictly be correct in this section.

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
 
 This is the binary format that the geometry is encoded in. The string 'WKB' to represent 
-[Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
-of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
+The string 'WKB', signifying [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
+of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB 
+representation. This means 3D coordinates are not supported in this version of GeoParquet, but we expect
+this to come in a future version.
+
+#### Coordinate Order
```
jorisvandenbossche, line 99:
```suggestion
#### Coordinate axis order
```

jorisvandenbossche (commented):
Thanks for the PR! Added a few more minor comments.

format-specs/geoparquet.md
alasarr: I prefer CRSes, it also uses in the previous paragraph

format-specs/geoparquet.md
alasarr: I think spatial reference is more popular and I don't expect someone to go there to get the WKT2 string, we already have an example in the readme for the WKT2 string.

If somebody else is happy with this change I'll add it, not a strong opinion here.

jorisvandenbossche: I know spatialreference.org is more popular, but unfortunately it is also horribly outdated. Its own home page says "Last update: 2013"

format-specs/geoparquet.md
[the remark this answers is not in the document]
jorisvandenbossche: Given that this is a new effort, I think there is something to be said for directly going with only WKT2:2019. 

I suppose the main disadvantage is that this new format is not yet as widely supported in libraries, making it more difficult to write compliant geoparquet files.

format-specs/geoparquet.md
alasarr: Ok, good point. Added.

format-specs/geoparquet.md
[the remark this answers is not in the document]
alasarr: We did a restriction with WKT2 by removing WKT1. The difference between WKT2 revisions is not really big, and it is well managed by proj4. Thus, I decided to go with all the WKT2 “flavors”, so it’s easier to write files in geoparquet.

I also want to avoid with this decision that people start creating geoparquet files non compliant with non supported WKT2 strings. It might happen that someone use a WKT2 2015 string without realizing of the mistake. If it happens with a large service and a huge amount of datasets out there, the only way to fix it is with another revision of geoparquet spec to add it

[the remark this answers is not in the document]
cholmes: I lean towards going just WKT2:2019 to start, and then see if there are implementations that aren't able to handle it, and then we can loosen it.

The main thing I want to be sure of is that EPSG:4326 has just *one* representation. So If 2015 vs 2019 for EPSG:4326 are different then I think it'd be really good to just do 2019. My thinking is that we shouldn't make implementations have to check for two different strings that mean the same thing (like if you have proj there).

Ideally we'd have some sort of validator that checks on if it's a 2019 string so we'd catch if a large service starts to use it wrong, so we wouldn't need a spec revision.

format-specs/geoparquet.md
cholmes: Yeah, I saw it - I think we can refer to that section. I just think we need to call it out here. Acknowledge it's a bit confusing. We say here 'use the crs for latitude/longitude', and then below we say 'but do it as longitude/latitude'. 

I suppose alternatively we don't actually mention lat/long or or long/lat here - we just that we recommend 4326 for widest interoperability.

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
 
 #### crs
 
-It is strongly recommended to use [EPSG:4326 (lat, long)](https://spatialreference.org/ref/epsg/4326/) for all data, so in most cases the value of the crs should be:
+The Coordinate Reference System (CRS) is a mandatory parameter for each geometry column defined in geoparquet format. 
+
+The CRS must be provided in [WKT](https://en.wikipedia.org/wiki/Well-known_text_representation_of_coordinate_reference_systems) version 2, also known as **WKT2**. WKT2 has several revisions, this specification supports the revisions from [2015](http://docs.opengeospatial.org/is/12-063r5/12-063r5.html) and [2019](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html): WKT2_2015, WKT2_2015_SIMPLIFIED, WKT2_2019, WKT_2019_SIMPLIFIED. 
+
+
+As the most common CRS for datasets is latitude/longitude, for the widest interoperability we recommend [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) for all data, so in most cases the value of the crs should be:
```
cholmes, line 63:
```suggestion
For the widest interoperability we recommend [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) for all data, as it is the most widely used coordinate reference system today, so unless data is stored in an alternate projection the crs should be:
```

format-specs/geoparquet.md
```diff
@@ -50,33 +50,55 @@ Each geometry column in the dataset must be included in the columns field above
...
-of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
+This is the binary format that the geometry is encoded in.
+The string 'WKB', signifying [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
+of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB 
+representation. This means 3D coordinates are not supported in this version of GeoParquet, but we expect
+this to come in a future version.
+
+#### Coordinate axis order
+
+The axis order of the coordinates in WKB stored in a geoparquet follows the de facto standard for axis order in WKB and is therefore always (x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS.
```
cholmes, line 101:
```suggestion
The axis order of the coordinates in WKB stored in a geoparquet follows the de facto standard for axis order in WKB and is therefore always 
(x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS. 
This follows the precedent of [GeoPackage](https://geopackage.org), see the [note in their spec](https://www.geopackage.org/spec130/#gpb_spec). 
```

cholmes:
> Now, if we want to discuss this, let's open a new issue for that (it's a hairy discussion with no easy answer, but just wanted to mention it here, being reminded about it seeing the ensemble datum in the WKT2 output). I think for this PR / for 0.1, the current recommendation is good enough (in the end, it's only a recommendation, and changing a recommendation is not a breaking change to the spec).

Yeah, it does feel weird to explain to users to use 4326 but ignore what it actually says. Fully agree on opening an issue and just shipping what we have in 0.1. I think it'd be good to get a wider audience to sound in on what we should do - I'm definitely going back and forth on it.

cholmes (approved):
I have two little commitable suggestions, but other than that this looks good, and I may be away from the computer for the rest of the day so giving my approval so this can merge.

format-specs/geoparquet.md
jorisvandenbossche: > I suppose alternatively we don't actually mention lat/long or or long/lat here - we just that we recommend 4326 for widest interoperability.

+1

format-specs/geoparquet.md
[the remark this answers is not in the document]
jorisvandenbossche: > The main thing I want to be sure of is that EPSG:4326 has just one representation

Note that because of the dynamic nature of "EPSG:4326" (ensemble datum that gets updated over time), the exact representation of it actually depends on the version of the EPSG database you are using. See eg the comment at https://github.com/opengeospatial/geoparquet/pull/24#discussion_r818778978, where our example parquet file already changed because of this

> Ideally we'd have some sort of validator that checks on if it's a 2019 string so we'd catch if a large service starts to use it wrong, so we wouldn't need a spec revision.

Just checking, and it seems that at least PROJ has functionality to check the dialect of a WKT string (although I don't directly see this exposed in pyproj)

jorisvandenbossche:
(small note: I pushed a commit to revert the changes to the example, to avoid conflicts with https://github.com/opengeospatial/geoparquet/pull/24)

format-specs/geoparquet.md
```diff
@@ -50,33 +50,57 @@ Each geometry column in the dataset must be included in the columns field above
...
+        UNIT["degree",0.0174532925199433],
+    USAGE[
+        SCOPE["Horizontal component of 3D system."],
+        AREA["World."],
+        BBOX[-90,-180,90,180]],
+    ID["EPSG",4326]]
 ```
 
-Data that is better served in particular projections can choose to use an alternate coordinate reference system.
+Due to the large number of CRSes available and the difficulty of implementing all of them, we expect that a number of implementations will at least start with only support a single CRS. To maximize interoperability we strongly recommend GeoParquet tool providers to always implement support for [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html).
```
jorisvandenbossche, line 88:
```suggestion
Due to the large number of CRSes available and the difficulty of implementing all of them, we expect that a number of implementations will at least start with only supporting a single CRS. To maximize interoperability we strongly recommend GeoParquet tool providers to always implement support for [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html). 
```

jorisvandenbossche (approved):

format-specs/geoparquet.md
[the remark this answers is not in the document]
alasarr: >Ideally we'd have some sort of validator that checks on if it's a 2019 string so we'd catch if a large service starts to use it wrong, so we wouldn't need a spec revision.

The issue is that we cannot guarantee every tool writing geoparquet file has a proper validator in place.

Anyways, it looks like I'm the only one supporting this idea 😄  and it's 0.1, so let's move on with WKT:2019 and we can revisit it in the future. 

Really appreciate your feedback here.

format-specs/geoparquet.md
alasarr: I see your point now 👍

jorisvandenbossche:
Thanks @alasarr !

format-specs/geoparquet.md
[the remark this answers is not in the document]
cholmes: Well, we don't need every tool writing geoparquet to have a proper validator in place. We just need one that the implementor creating a tool can use to check their output to ensure it aligns, and/or that data users can use to then give feedback to the implementor.

And more than happy to revisit - as with other discussions I think it's easier to come back and make things looser than it is to try to tighten later.

PR #27, opened 2022-03-04
"edges" field spec definition for version 0.1

### LINKED ISSUE #9: Store information about planar vs spherical coordinates (eg geodesic=true/false)
This has come up in several places (eg most recently in https://github.com/opengeospatial/cdw-geo/issues/3#issuecomment-1047795817), and brought up in a previous meeting.

Geospatial analytical systems can interpret / treat geometries' coordinates as planar or spherical. For example GEOS considers everything as planar coordinates (and thus also GeoPandas, or R's `sf` or PostGIS when using GEOS). Other libraries can handle spherical coordinates, such R's `s2` package (which is now used as default in `sf` when having geographical coordinates) or BigQuery's Geography functionality (I _think_ both based on Google's `s2geometry`). PostGIS also differentiates between a geometry and geography type. 

Once you deal with spherical coordinates, you also have to deal with the edges of geometries. A geometry can be valid (i.e. no intersecting edges) when interpreting the coordinates as planar, but could be invalid when interpreting the geometries as spherical. 
And that's where the IO aspect comes into the picture. When reading in data (and working with spherical coordinates), you can either 1) assume the edges are already valid as spherical coordinates, or 2) do a conversion of planar edge to spherical edge. 

For example, BigQuery assumes spherical edges when reading in from WKT (with the `planar=TRUE` option in `ST_GEOGFROMTEXT` to override this default), but planar edges when parsing GeoJSON (see https://cloud.google.com/bigquery/docs/geospatial-data#coordinate_systems_and_edges).

Quoting @paleolimbot:

> There is currently no way to communicate in any file "this was exported from BigQuery Geography or S2 so you can import it there again without tessellating all the edges again" (e.g., use [planar = true when importing to BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_geogfromtext_signature2)).

So it would be useful to store this information about the edges _in the file metadata_, instead of having the user of the data to know this and specify it as an option while reading the data.

The concrete proposal would be to have an additional column metadata field to indicate this. I think a boolean flag is fine for this, and possible names are `"geodesic": true/false` or `"planar": true/false`.

---

<sub>Note: I am no expert on this front (GeoPandas is, for now, still only using GEOS and thus planar coordinates, so I don't have much experience with handling spherical coordinates). So please correct me if anything in the above isn't fully correct :) </sub>

Jesus89:
This PR aims to cover the definition of "edges" field for version 0.1.

Closes #9


cholmes (approved):
Looks good. Put one small change to make it clear there are just three options, and also to change srid to crs.

At some point I think it'd be good to explain more what this is, some sort of primer on why it matters, how it interacts with the coordinate reference system. Probably doesn't need to be in the spec itself.

format-specs/geoparquet.md
```diff
@@ -79,6 +80,15 @@ This is the binary format that the geometry is encoded in. The string 'WKB' to r
 [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
 of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
 
+#### edges
+
+This attribute indicates the coordinate system to be applied when interpreting the edges. Available values are:
+- planar: use flat cartesian coordinate system.
+- spherical: use spherical coordinate system and radius derived from the spheroid defined by the SRID.
+- spheroid: use an spheroid or ellipsoid defined by the SRID.
```
paleolimbot, line 88: I like the use of 'spheroid', since that's also how it shows up in the WKT2. I think that "spherical" and "spheriod" should match, though (i.e., "sphere" and "spheroid" OR "spherical" and "spheroidal").

paleolimbot (commented):
This looks great!

format-specs/geoparquet.md
cholmes: +1 - just changed my suggestion to be spheroidal (since I think if we go sphere and spheroid then we also should say 'plane' instead of planar?)

jorisvandenbossche: At some points we mentioned "let's leave out spheriod / ellipsoid (for 0.1) until we actually have a use case / someone asking for it" during the meeting, but I don't know if we made a clear decision on that.

Jesus89: Agree with the new naming suggestion @paleolimbot @cholmes.

Regarding the inclusion of `spheroidal` in the 0.1 @jorisvandenbossche: we didn't make a final decision on that. I included it for completeness and because it's an informative attribute, but if we all agree on leaving it out of version 0.1, I will apply the changes in the branch.

jorisvandenbossche: > As a note, BigQuery already supports spheroidal precision in the GIS functions, for example [ST_DISTANCE](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_distance).

That is for an operation, and not for IO. When reading data, it seems they don't make a distinction between spherical or spheroidal (see https://cloud.google.com/bigquery/docs/geospatial-data#coordinate_systems_and_edges, which mentions a planar=True/False flag for [ST_GEOGFROMTEXT](https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_geogfromtext_signature2))

format-specs/geoparquet.md
```diff
@@ -79,6 +80,15 @@ This is the binary format that the geometry is encoded in. The string 'WKB' to r
 [Well Known Binary](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary) is the only current option, but future versions
 of the spec may support alternative encodings. This should be the ["standard"](https://libgeos.org/specifications/wkb/#standard-wkb) WKB representation.
 
+#### edges
+
+This attribute indicates the coordinate system to be applied when interpreting the edges. Available values are:
```
jorisvandenbossche, line 85: I find the current use of "coordinate system" in "indicates the coordinate system to be applied when interpreting .." a bit confusing, since that word is closely related to coordinate reference system, and this field is independent from the CRS. 
I know that strictly speaking this is correct (since it's about using a cartesian vs spherical coordinate system), but since both can be used while using a geographical coordinate system, I think we might need to explain this a bit more to avoid confusion.

A first attempt:

> This attribute indicates how to interpret the edges of the geometries: whether the line between two points is a straight Cartesian line or the shortest line on the sphere (geodesic line).

cholmes: +1 on clarifying that. Your attempt looks good to me.

format-specs/geoparquet.md
cholmes: I think we should leave it out for 0.1. It's easy to add it in when there's a use case. But it's hard to remove it, since we can't be sure that no one used it and relies on it after we put this out. I think we should open an issue for it, show how we're planning to represent it and ask for feedback. Then hopefully people will easily google it and can comment on their use case if they're looking for it.

Jesus89: Agree with both. Regarding IO BigQuery distinguishes between planar (GeoJSON or WKT+planar flag) or spherical (WKT by default).

I think this opens the debate of using boolean vs string, so planar flag vs edges enum.

format-specs/geoparquet.md
Jesus89: +1

alasarr (approved):
Looks great to me

PR #32, opened 2022-03-08
Added parquet-wasm to libraries

cholmes:
Added an 'in-progress' section and added https://github.com/kylebarron/parquet-wasm

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,3 +86,7 @@ Currently known libraries that can read and write GeoParquet files:
 
 * [GeoPandas](https://geopandas.org/en/stable/docs/user_guide/io.html#apache-parquet-and-feather-file-formats) (Python)
 * [sfarrow](https://wcjochem.github.io/sfarrow/index.html) (R)
+
+In progress (beta) libraries:
+
+* [Parquet-WASM](https://github.com/kylebarron/parquet-wasm) (Rust-based WebAssembly)


kylebarron (approved):
Looks, good. Thanks!

jorisvandenbossche:
Cool to see better in-browser tools for Parquet!

Just wondering: I _think_ it is currently more a general Parquet reader/writer, how do you see specific support for the GeoParquet spec fitting in? (I am not familiar with the JS space, so no idea if there are specific geospatial tools that would be useful to connect to when reading/writing the data)

kylebarron:
Yes, that project is a general Parquet reader and writer. 

In the [Vis.gl org](https://github.com/visgl?type=source), we have a repo called [loaders.gl](https://github.com/visgl/loaders.gl), which includes parsers (and sometimes writers) for lots of different file formats, including geospatial ones like Shapefile, Flatgeobuf, GeoPackage, MVT, WKB, WKT, KML, GPX and some non-geo formats. One of the goals of loaders.gl is to present a unified output format so that libraries like [deck.gl](https://github.com/visgl/deck.gl) can consume data from any source seamlessly. (For now, this means often parsing data to GeoJSON to pass to deck.gl, but ideally in the future loaders.gl would add GeoArrow as a choice of output format).

So ideally in the future you'll be able to do something like and it _should_ \*just work\* 😄 .

```js
import {GeoJsonLayer} from '@deck.gl/geo-layers';
import {GeoParquetLoader} from '@loaders.gl/parquet';

const dataUrl = 'https://example.com/data.parquet';
const layers = new GeoJsonLayer({
  id: 'parquetData',
  data: dataUrl,
  loaders: [GeoParquetLoader]
})

return <DeckGL layers={layers} />
```

So to get back to your point, I'm not sure what if anything is worthwhile to include in the list here. I'm just starting to try and integrate `parquet-wasm` with loaders.gl, so full support to deck.gl is still a little ways off.

jorisvandenbossche:
@kylebarron thanks for that explanation!

> So to get back to your point, I'm not sure what if anything is worthwhile to include in the list here. I'm just starting to try and integrate `parquet-wasm` with loaders.gl, so full support to deck.gl is still a little ways off.

Eventually, I suppose it would be mainly `loaders.gl` to list here? (with a mention that it can be used in deck.gl) For now I am also fine with mentioning `parquet-wasm` as a way to read Parquet data in the browser, we can maybe add the disclaimer that you still need to parse the WKB and metadata yourself for now.

alasarr (approved):

kylebarron:
> disclaimer that you still need to parse the WKB and metadata yourself for now.

I think part of the issue here is that there isn't really a canonical geometry format on the web, or at least not a performant one. In Python, a program that "reads GeoParquet" could reasonably be expected to load into GEOS objects. But the closest thing to a canonical representation on the web is probably GeoJSON, but parsing to that would lose a lot of the performance benefits of Parquet/Arrow.

If GeoArrow becomes a canonical geometry format, and GeoArrow and GeoParquet share the same geometry format, then the beauty is that any Parquet reader becomes a GeoParquet-specific reader.

jorisvandenbossche:
@kylebarron that's a good point. So then we certainly don't have to wait on any hypothetical future library to include a mention of this. I think it is still worth to add a brief mention that this is a pure parquet reader and that there is not special handling of the metadata (eg for writing you will need to write up those metadata yourself).

felixpalmer:
> But the closest thing to a canonical representation on the web is probably GeoJSON, but parsing to that would lose a lot of the performance benefits of Parquet/Arrow.

Specifically in the case of deck.gl and the `GeoJsonLayer` we already have a binary format that more closely maps to what deck.gl needs to send the data off for rendering to the GPU. It is described here: https://loaders.gl/modules/gis/docs/api-reference/geojson-to-binary

This works, however the binary format is difficult for humans to work with and it would be much better if we could use GeoArrow as this provides user-friendly methods to filter and manipulate the data.

In the short term it might be worth trying to map the loaders.gl output from the `GeoParquetLoader` to this binary format so we can get an idea about the performance and have a way to visualize data with the existing deck.gl library in the browser

cholmes:
Going to close this, since I think this list should be tools/libraries that will get users going with geoparquet. Plus I'm moving the list more to the website. But deck.gl / loaders.gl support for geoparquet would be amazing.

PR #36, opened 2022-03-09
De-nest examples directory

jorisvandenbossche:
Now this repo is only about "geoparquet"


TomAugspurger (approved):
I might re-nest it again in the future ("basic", "spheroidal", ...) but this is better for now :)

PR #45, opened 2022-03-17
Expand WKB encoding to ISO WKB to support 3D geometries

### LINKED ISSUE #18: Support 3D coordinates
For this we'll need to decide on [EWKB](https://libgeos.org/specifications/wkb/#extended-wkb) or 
[ISO WKB](https://libgeos.org/specifications/wkb/#iso-wkb)

From what everyone could tell the core WKB is the same with both, so we're starting with that. But we should have experts sound in on what we should use. 

Initial research shows ISO WKB is more of a 'standard', as EWKB has not been formalized in a document (though it has extensive support). With our CRS field we don't actually need the SRID part of EWKB.

jorisvandenbossche:
Closes #18


cholmes (approved):
Sorry for the slow response - looks good.

cholmes:
Merging this in, as it's been open for awhile so has had time for any objections.

Issue #35, opened 2022-03-09
How to deal with dynamic CRS or CRS with ensemble datums (such as EPSG:4326)?


jorisvandenbossche:
From https://github.com/opengeospatial/geoparquet/pull/25#issuecomment-1059016020. The spec has a required `crs` field that stores a WKT2:2019 string representation of the Coordinate Reference System.

We currently recommend using EPSG:4326 for the widest interoperability of the written files. However, this is a dynamic CRS, and in addition uses an ensemble datum. See https://gdal.org/user/coordinate_epoch.html for some context. In summary, when using coordinates with a dynamic CRS, you also need to know the point in time of the observation to know the exact location.

Some discussion topics related to this:

- How do we deal with a dynamic CRS? We should probably give the option to include the "coordinate epoch" in the metadata (the point in time at which the coordinate is valid)
  - This coordinate epoch is _not_ part of the actual CRS definition, so I think the most straightforward option is to specify an additional (optional) `"epoch"` field in the column metadata (next to `"crs"`) that holds the epoch value as a decimal year (eg 2021.3).
  - This means we would only support a constant epoch _per file_. This is in line with the [initial support for other formats in GDAL](https://gdal.org/user/coordinate_epoch.html#support-in-raster-and-vector-formats), and we can always later think about expanding this (eg additional field in the parquet file that has a epoch per geometry, or per vertex)

paleolimbot:
It's my understanding that there is a place for the coordinate epoch in WKT2:2019 (i.e., by defining a CRS with one of the realizations as the GDAL page described). If that is the case, I would suggest leaving the coordinate epoch representation to the WKT2 representation in the crs field (but that should be checked!). In theory you can also put that in the coordinates themselves (e.g., via an "m" coordinate) and define a spatiotemporal CRS. I'd argue that anybody that cares about sub-meter accuracy will also be using a projected CRS, though.

paleolimbot:
I stand corrected...sorry for the indirection! It seems that there's a better chance it will propagate through an operation if kept in the metadata under `epoch` since it might get dropped if read into PROJ and then re-rendered as WKT?

jorisvandenbossche:
[...]

PR #49, opened 2022-03-25
Add "epoch" field to optionally specify the coordinate epoch for a dynamic CRS

jorisvandenbossche:
See first bullet point from https://github.com/opengeospatial/geoparquet/issues/35 (and the following discussion).

This borrows a bit of text from https://gdal.org/user/coordinate_epoch.html (cc @rouault)

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -52,6 +52,7 @@ Each geometry column in the dataset must be included in the columns field above 
 | Field Name |                               Type                                      |                                                                   Description                                                                     |
 | ---------- | ----------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
 | crs       | string   | **REQUIRED** [WKT2](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html) string representing the Coordinate Reference System (CRS) of the geometry.  |
+| epoch    | double | **OPTIONAL** Coordinate epoch in case of a dynamic CRS, expressed as a decimal year.  |
 | encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported. |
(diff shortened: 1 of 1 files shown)


cholmes (approved):
Looks good. Though does 4326 need an epoch? If so we should add what people should put for the epoch if they are using our crs recommendation for interoperability.

rouault:
> Though does 4326 need an epoch?

that's a question that could bring geodesists to disagree for hours.
Strictly speaking EPSG:4326 relies on the EPSG:6326 datum ensemble (https://epsg.org/datum_6326/World-Geodetic-System-1984-ensemble.html), that has an intrinsic inaccuracy of 2 meters, due to the oldest realizations like Transit being quite different from the latest ones, that disagree between them only at the centimer level. So for such a vaguely qualified object, a coordinate epoch doesn't make much sense.
However ... in many cases people may use the vague EPSG:4326 instead of using one of the latest realizations of WGS 84 or ITRFxxxx, but for a number of reasons (like people only knowing EPSG:4326 and freaking out when they see something different) they use EPSG:4326 instead of the proper code. In that case, qualifying the CRS with the epoch could be useful for later processing, where people would say "oh, they actually meant ITRF2014@2022.3".
There are even cases for a CRS/datum that is generally considered static (and for which a coordinate epoch normally doesn't make sense, ), like NAD83(2011), where adding a coordinate epoch might be useful, since there are high precision deformation models that can be used.

cholmes:
Gotcha, those explanations make sense. And yeah, I was thinking something like 'if you use 4326 then do X for epoch', but if it doesn't really matter then that seems fine to not.

> (in general we might want to reconsider the recommendation of EPSG:4326 as I mentioned in https://github.com/opengeospatial/geoparquet/issues/35 (the second bullet point), but let's leave that discussion for in the issue)

Let's actually break that out into its own issue. I've been having doubts about it as well.

alasarr (approved):

PR #51, opened 2022-03-25
Add optional "geometry_type" field per column

### LINKED ISSUE #41: Advertizing geometry field "type" in Column metadata ?
A common use case if for a geometry column to hold a single geometry type (Point, LineString, Polygon, ...) for all its records. It could be good to have an optional "type" field name under https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#column-metadata to capture that when it is known.

This would help for example conversion between GeoParquet and GIS formats (shapefiles, geopackage) that have typically this information in their metadata.

Values for type could be the ones accepted by GeoJSON: Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon, GeometryCollection  (that would be extended to  CircularString, CompoundCurve, CurvePolygon, MultiCurve, MultiSurface, PolyhedralSurface, TIN if we support ISO WKB,  and with Z, M or ZM suffixes for other dimensionalities)

What to do when there is mixed geometry types ?
- do not set "type"
- set "type": "mixed"
- set "type": array of values, e.g. [ "Polygon", "MultiPolygon" ]  (this one would be typical when converting from shapefiles where the polygonal type can hold both polygons and multipolygons )

jorisvandenbossche:
Closes #41


cholmes:
@jorisvandenbossche - what's needed to finalize this PR? It's still marked 'draft'. Would be great to get this in for 0.2.0.

Jesus89:
In this old version of "GeoParquet" ([link](https://github.com/darcy-r/geoparquet-python/blob/master/geoparquet/__init__.py#L77)) there is a similar approach for storing the types. In that case, it uses `geometry_types` in plural so it always will be an array.

I think it's fine having both `string` and `array` but wanted to share the info, just in case.

BTW, I miss the implementation of the field in the example. Let me know if you want me to add it.

jorisvandenbossche:
> what's needed to finalize this PR?

I think mostly some more feedback on what I actually added here / the open questions in https://github.com/opengeospatial/geoparquet/issues/41

cholmes:
> I think mostly some more feedback on what I actually added here / the open questions in https://github.com/opengeospatial/geoparquet/issues/41

Ah, got it. I see - this currently does have it as an optional field and Unknown as an option. I do agree with [your comment](https://github.com/opengeospatial/geoparquet/issues/41#issuecomment-1079172309) that we should pick one so there's not two ways to say unknown. I lean towards required and allow unknown. Follows flatgeobuf.

alasarr (approved):
👏 Super happy to see this new field. Thank you so much @jorisvandenbossche and @rouault! I've faced this last week and a few minutes ago with multiple use cases where it's required. 

Today: The tilers at CARTP need to guess the type of geometry, right now we need to add a query over the data of the user to determine the type of geometry (using samples and some heuristics... 😢 ). It's something not only for GeoParquet, it should be added as a feature in the data warehouses, i.e., given a geom column of a table have a way to get the geometry type from some metadata without exploring the data.

Jesus89 (approved):

format-specs/geoparquet.md
```diff
@@ -53,6 +53,7 @@ Each geometry column in the dataset must be included in the columns field above
 | ---------- | ----------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------- |
 | crs       | string   | **REQUIRED** [WKT2](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html) string representing the Coordinate Reference System (CRS) of the geometry.  |
 | encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported. |
+| geometry_type | string or \[string] | **OPTIONAL** If known, the geometry type(s) of all geometries.  |
```
cholmes, line 56:
```suggestion
| geometry_type | string or \[string] | **REQUIRED** The geometry type(s) of all geometries, or "Unknown" if they are not known.  |
```

format-specs/geoparquet.md
```diff
@@ -104,6 +105,27 @@ The axis order of the coordinates in WKB stored in a geoparquet follows the de f
 (x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS. 
 This follows the precedent of [GeoPackage](https://geopackage.org), see the [note in their spec](https://www.geopackage.org/spec130/#gpb_spec). 
 
+#### geometry_type
+
+This optional field captures the geometry type(s) of the geometries in the
```
cholmes, line 110:
```suggestion
This field captures the geometry type(s) of the geometries in the
```

format-specs/geoparquet.md
```diff
@@ -104,6 +105,27 @@ The axis order of the coordinates in WKB stored in a geoparquet follows the de f
...
+
+In addition, the following rules are used:
+
+- In case of 3D geometries, a " Z" suffix gets added (e.g. "Point Z").
+- The value can be a single string or an array of strings in case multiple
+  geometry types are present (e.g. ["Polygon", "MultiPolygon"]).
+- Additionally the value "Unknown" is accepted to explicitly signal that the
+  geometry type is not known.
+
+If specified, it is expected that this field is strictly correct. For
```
cholmes, line 123:
```suggestion
It is expected that this field is strictly correct. For
```

cholmes (commented):
I added some edits to make it so the field is required, but unknown is allowed. 

I'm also ok if we make it optional and specify that leaving it out is the same as 'unknown', but prefer required & unknown option, to nudge implementors to thinking about it.

alasarr:
As @jorisvandenbossche already [agreed](https://github.com/opengeospatial/geoparquet/issues/41#issuecomment-1101535577) on the idea of required & unknown and @cholmes too, I'm merging.

Issue #46, opened 2022-03-23
Define polygon orientation rules


mentin:
I think the standard should define polygon orientation.

### 1. Spherical edges case
With spherical edges on sphere, there is an ambiguity in polygon definition, if the system allows polygons larger than hemisphere.

A sequence of vertices that define a polygon boundary can define either polygon to the left of that line, or to the right of the line. E.g. global coastal line can define either continents or oceans. Systems that support polygons larger than hemisphere usually use orientation rule to solve this ambiguity. E.g. [MS SQL](https://docs.microsoft.com/en-us/sql/relational-databases/spatial/spatial-data-types-overview?view=sql-server-ver15#orientation-of-spatial-data), [Google BigQuery](https://cloud.google.com/bigquery/docs/geospatial-data#polygon_orientation) interpret the side to the left of the line as the content of the ring.

### 2. Planar edges case

Planar case does not have such ambiguity, but it is still good idea to have specific rule.

E.g. [GeoJson RFC](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6) defines a rule consistent with the rule above:
```
   o  Polygon rings MUST follow the right-hand rule for orientation
      (counterclockwise external rings, clockwise internal rings).
```

cholmes:
Very good call, thanks @mentin! If you'd like to contribute directly to the spec feel free to open a pull request. If not I'm sure someone will be able to do it - hoping we can get it in before 0.2. It'll be good to have this clear.

alasarr:
Nice catch @mentin! I can help to open a PR to include this, just let me know.

According to #2. As @rouault mentions, the current format we use for encoding geometries (WKB) doesn't impose that constraint, so my preference would be to not include it unless we have a specific reason like #1. Not a strong opinion here, just want to keep as simpler as possible at the beginning to increase adoption. Does it make sense for you @mentin?

mentin:
Yes, it makes sense to only require it for spherical case. Our (BigQuery) experience with GeoJson was that despite the RFC requiring specific orientation, we frequently see opposite orientation too, so we ignore it when reading GeoJson files.

felixpalmer:
I think it is instructive to observe what happened with GeoJSON here. The winding order was omitted from the original spec and then later added. As a result, spec-compliant GeoJSON does not require a winding check and potential re-winding, but in practice clients need to deal with old GeoJSON out there so all perform the winding check.

[...]

PR #59, opened 2022-04-11
Add section on winding order

felixpalmer:


alasarr (approved):

cholmes:
I'm out of my depth here, don't think I've ever dug into winding order. Hoping @tschaub or @hobu sounds in on #46 from the geojson perspective.

cholmes:
I'm out of my depth here, don't think I've ever dug into winding order. Hoping @tschaub or @hobu sounds in on #46 from the geojson perspective.

Jesus89:
From this https://github.com/opengeospatial/geoparquet/issues/46#issue-1178648653, this covers the `planar` edges case, but not the `spherical` edges case. I would make this explicit in the spec. Regarding the spherical case, we can leave it undefined or with an explicit definition too.

alasarr:
>From this https://github.com/opengeospatial/geoparquet/issues/46#issue-1178648653, this covers the planar edges case, but not the spherical edges case. I would make this explicit in the spec. Regarding the spherical case, we can leave it undefined or with an explicit definition too.

I think @felixpalmer proposes to force this winding order for both cases

cholmes:
> I think @felixpalmer proposes to force this winding order for both cases

Yes, but I *think* we also need to specify the orientation rule. It probably makes sense to follow [MS SQL](https://docs.microsoft.com/en-us/sql/relational-databases/spatial/spatial-data-types-overview?view=sql-server-ver15#orientation-of-spatial-data), and [Google BigQuery](https://cloud.google.com/bigquery/docs/geospatial-data#polygon_orientation) and 'interpret the side to the left of the line as the content of the ring'.

jorisvandenbossche:
I commented on the issue for the more general discussion (https://github.com/opengeospatial/geoparquet/issues/46).

But specific comment on the text in this PR: can we avoid using the term "right hand rule"? I know the GeoJSON spec uses this, but that term is very confusing, as other sources use "left hand" to specify the same .. (eg the cited bigquery and MS SQL sources)

alasarr:
Thanks @jorisvandenbossche and @cholmes. @felixpalmer I think makes sense to specify that to be more explicit.

format-specs/geoparquet.md
```diff
@@ -105,6 +105,10 @@ The axis order of the coordinates in WKB stored in a geoparquet follows the de f
 (x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS. 
 This follows the precedent of [GeoPackage](https://geopackage.org), see the [note in their spec](https://www.geopackage.org/spec130/#gpb_spec). 
 
+#### Polygon winding
+
+The winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).
```
tschaub, line 110:
```suggestion
The winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
```

tschaub:
Late to the party, but I agree that this is good to specify.  I think mentioning the "right-hand rule" is ok (as some people may know it), but that it also makes sense to add some more detail (something like "the interior of the polygon is to the left of the boundary as you traverse vertices in order").

Inevitably, consumers of the data will do orientation checks before rendering, topology operations, etc. (as the data gets passed through structures that don't keep track of or require a specific orientation), but I still think it improves things to have the format specify how polygon rings are to be interpreted.

format-specs/geoparquet.md
rouault: Perhaps it should also be mentioned that this is the opposite convention of a little-known format called Shapefile :-)

alasarr:
@tschaub this is one of the latest PRs of v02.0. I've added your commit suggestion. Could you add your stamp here?

cholmes (approved):

felixpalmer:
I agree on the clarification here, it makes it more accessible to wider reader base

PR #58, opened 2022-04-05
Add validator script for Python based on JSON Schema

Jesus89:
Implements https://github.com/opengeospatial/geoparquet/issues/23.

This PR adds a draft for a basic validator using Python. It checks the metadata of a geoparquet file using JSON Schema, but it can be extended to include specific custom validations.

**Example**

Try to validate this wrong metadata

```py
metadata = {
    "version": "0.x.0",
    "primary_column": "geom",
    "columns": {
        "geometry": {
            "encoding": "WKT",
            "edges": "",
            "bbox": [180, 90, 200, -90],
        },
    },
}
```

```
$ python3 validator/validate.py examples/example-wrong.parquet
Validating file...
- [ERROR] $.version: '0.x.0' does not match '^0\\.1\\.[0-9]+$'
          INFO: The version of the geoparquet metadata standard used when writing.
- [ERROR] $.columns.geometry.encoding: 'WKT' is not one of ['WKB']
          INFO: Name of the geometry encoding format. Currently only 'WKB' is supported.
- [ERROR] $.columns.geometry.edges: '' is not one of ['planar', 'spherical']
          INFO: Name of the coordinate system for the edges. Must be one of 'planar' or 'spherical'. The default value is 'planar'.
- [ERROR] $.columns.geometry.bbox[2]: 200 is greater than the maximum of 180
          INFO: The eastmost constant longitude line that bounds the rectangle (xmax).
- [ERROR] $.columns.geometry: 'crs' is a required property
- [ERROR] $.primary_column: must be in $.columns
This is an invalid GeoParquet file.
```

The output for a correct GeoParquet file

```
$ python3 validator/validate.py examples/example.parquet
Validating file...
This is a valid GeoParquet file.
```

PR #62, opened 2022-04-12
Validator updates

kylebarron:
### Change list

- Add support for remote files via `http://`, `https://`, `s3://`, and `gs://` protocols.
- Document in the `README.md` that `s3fs` needs to be installed to load `s3://` files and `gcsfs` needs to be installed to load `gs://` files.
- Specify in the `environment.yml` file that jsonschema v4 is required. The error that @cholmes hit [here](https://github.com/opengeospatial/geoparquet/pull/58#issuecomment-1095529120) is likely because he had jsonschema v3 installed.
- Raise error if the Parquet file metadata has no `geo` key
- Don't suggest that users call `./validate.py`. Especially in conjunction with `#!/usr/bin/python3`, that would've always used the _system_ Python executable, which is both bad practice and would likely lead to confusion why Python packages weren't being found. Instead, suggesting `python validate.py` ensures that the python in the user's `PATH` will be used.
- Change the shebang from `#!/usr/bin/python3` to `#!/usr/bin/env python3`, which ensures that if the user _does_ call `./validate.py`, then most often the `python` in the user's `PATH` will be found, instead of always using the system Python.
- Use 2-space indenting for the `schema.json` file in line with the [STAC spec's schema files](https://github.com/radiantearth/stac-spec/blob/master/item-spec/json-schema/item.json) for readability. I can back this out of this PR if preferred. Ideally we'll add a lint step to format the schema file.

### DIFF
--- a/validator/README.md
+++ b/validator/README.md
@@ -5,5 +5,12 @@ Based on [JSON Schema](https://json-schema.org/).
 ## Python
 
 ```
-./validate.py ../examples/example.parquet
+python validate.py ../examples/example.parquet
+python validate.py https://storage.googleapis.com/open-geodata/linz-examples/nz-buildings-outlines.parquet
 ```
+
+The validator also supports remote files.
+
+- `http://` or `https://`: no further configuration is needed.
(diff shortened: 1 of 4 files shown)


Jesus89:
Thanks! I'll merge the PR and integrate with my changes :+1:

Issue #65, opened 2022-04-12
Store JSON copy of metadata next to example Parquet files


kylebarron:
Making a separate issue to follow up on https://github.com/opengeospatial/geoparquet/pull/60#discussion_r848664664

I propose that for our examples we include a JSON file next to each Parquet file that mirrors its metadata. For example:

```bash
example1.parquet
# plain text schema that matches what's stored in example1.parquet
example1_schema.json

example2.parquet
# plain text schema that matches what's stored in example2.parquet
example2_schema.json
```

Benefits:

- Should make it easier for people browsing the spec to see schema examples without needing to open up a file.
- Git diffs should be easier to read and understand. Note that this requires that JSON files are always formatted in the same way, which we could enforce via a CI lint step.

We should also have a CI step that verifies the plain text schema in `example1_schema.json` actually matches what's actually stored in `example1.parquet`, and ideally that the schema passes validation (see #64 and #58)

cholmes:
+1 - seems like it'll make it much easier for people to see what's going on. And definitely +1 on CI to check formatting and that the plain text schema matches its example.

PR #68, opened 2022-04-15
Script to store JSON copy of metadata next to example Parquet files

kylebarron:
### Changes

- New `scripts/` folder to hold scripts that we don't expect to be useful outside of the repo.
- Python script to read all the Parquet files in the `examples/` directory and store a JSON copy of the schema metadata (note, not the entire schema, just the metadata). Using `indent=2` and `sort_keys=True` to ensure JSON files are the same each time.
- [Poetry](https://github.com/python-poetry/poetry) dependency specification and lockfile so that scripts are reproducible. I'm a big fan of lockfiles to ensure reproducibility, but if people are opposed to this, we could instruct people to use a pip command.
- CI config to run the script on each commit and PR and verify that the checked-in schema is up to date.

To run the script:

```
cd scripts && poetry install && poetry run python update_example_schemas.py 
```

Closes https://github.com/opengeospatial/geoparquet/issues/65

### DIFF
--- a/examples/example_metadata.json
+++ b/examples/example_metadata.json
@@ -1,0 +1,19 @@
+{
+  "geo": {
+    "columns": {
+      "geometry": {
+        "bbox": [
+          -180.0,
+          -90.0,
+          180.0,
+          83.6451
+        ],
+        "crs": "GEOGCRS[\"WGS 84\",ENSEMBLE[\"World Geodetic System 1984 ensemble\",MEMBER[\"World Geodetic System 1984 (Transit)\"],MEMBER[\"World Geodetic System 1984 (G730)\"],MEMBER[\"World Geodetic System 1984 (G873)\"],MEMBER[\"World Geodetic System 1984 (G1150)\"],MEMBER[\"World Geodetic System 1984 (G1674)\"],MEMBER[\"World Geodetic System 1984 (G1762)\"],MEMBER[\"World Geodetic System 1984 (G2139)\"],ELLIPSOID[\"WGS 84\",6378137,298.257223563],ENSEMBLEACCURACY[2.0]],CS[ellipsoidal,2],AXIS[\"geodetic latitude (Lat)\",north],AXIS[\"geodetic longitude (Lon)\",east],UNIT[\"degree\",0.0174532925199433],USAGE[SCOPE[\"Horizontal component of 3D system.\"],AREA[\"World.\"],BBOX[-90,-180,90,180]],ID[\"EPSG\",4326]]",
+        "edges": "planar",
+        "encoding": "WKB"
+      }
+    },
+    "primary_column": "geometry",
+    "version": "0.1.0"
+  }
+}
--- a/scripts/poetry.lock
+++ b/scripts/poetry.lock
@@ -1,0 +1,79 @@
+[[package]]
+name = "numpy"
+version = "1.22.3"
+description = "NumPy is the fundamental package for array computing with Python."
+category = "main"
+optional = false
+python-versions = ">=3.8"
+
+[[package]]
+name = "pyarrow"
+version = "7.0.0"
+description = "Python library for Apache Arrow"
+category = "main"
+optional = false
+python-versions = ">=3.7"
+
+[package.dependencies]
+numpy = ">=1.16.6"
+
+[metadata]
+lock-version = "1.1"
+python-versions = "^3.8"
+content-hash = "88e7b3e944bfcbba273f809c68223fe4fd1e90c8cf12a59eedaf64845df668c8"
+
+[metadata.files]
+numpy = [
+    {file = "numpy-1.22.3-cp310-cp310-macosx_10_14_x86_64.whl", hash = "sha256:92bfa69cfbdf7dfc3040978ad09a48091143cffb778ec3b03fa170c494118d75"},
+    {file = "numpy-1.22.3-cp310-cp310-macosx_11_0_arm64.whl", hash = "sha256:8251ed96f38b47b4295b1ae51631de7ffa8260b5b087808ef09a39a9d66c97ab"},
+    {file = "numpy-1.22.3-cp310-cp310-manylinux_2_17_aarch64.manylinux2014_aarch64.whl", hash = "sha256:48a3aecd3b997bf452a2dedb11f4e79bc5bfd21a1d4cc760e703c31d57c84b3e"},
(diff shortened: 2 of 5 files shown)


scripts/update_example_schemas.py
```diff
@@ -0,0 +1,55 @@
...
+    new_dict = {}
+    for key, val in d.items():
+        new_key = key.decode() if isinstance(key, bytes) else key
+        new_val = val.decode() if isinstance(val, bytes) else val
+        if new_val.lstrip().startswith("{"):
+            new_val = json.loads(new_val)
+
+        new_dict[new_key] = new_val
+
+    return new_dict
```
kylebarron, line 45: Note that if we only care about the metadata in the `geo` key, this can be simplified to
```
return json.loads(d[b'geo'])
```
as in:
https://github.com/opengeospatial/geoparquet/blob/8a996a3dd477491988f7b8451b9b5de509c32c24/examples/example.py#L9
But I wasn't sure if there would be a use case where we wanted to have other file-level metadata that also gets copied into the JSON file.

cholmes: Seems useful to me to try to print out other metadata that might be in there. Would help people catch if they accidentally didn't use the geo key, or use it right.

scripts/update_example_schemas.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def find_root_dir() -> Path:
+    """Find root geoparquet directory
+
+    This is a more robust approach than assuming __file__ refers to a specific place
+    """
+    folder = Path(__file__).resolve()
+
+    while folder.name != "geoparquet":
```
cholmes, line 15: Note this fails if the user checked out the repo with a name other than geoparquet. Probably not all that likely to happen, but I happened to have it named differently so it didn't find it. Was easy to fix.

cholmes (approved):
This worked well enough for me that I'm happy to approve - I tested it with the nz-buildings example, seemed to get the new JSON in the examples directory. 

But I do think it'd be good to have a simple readme, telling users what they need to do in order to use the script. As mr. naive python user I didn't know what to do with poetry.lock or pyproject.toml. I just ran `python update_example_schemas.py` and it worked. Maybe that is all I need to do, but a readme telling me to do so would make me feel more confident.

scripts/poetry.lock
```diff
@@ -0,0 +1,79 @@
+[[package]]
```
Jesus89, line 1: I wouldn't include a lock file for an internal tool. But no strong opinion on that.

scripts/pyproject.toml
```diff
@@ -0,0 +1,16 @@
+[tool.poetry]
```
Jesus89, line 1: It's fine to use poetry although I've not used it a lot, we should standardize the tools across the repo

examples/example_metadata.json
```diff
@@ -0,0 +1,19 @@
+{
```
Jesus89, line 1: I would encode just the "geo" metadata because this could confuse the reader about the internal structure of metadata.

The real structure is `b"geo": "{<geometadata>}"` instead of `"geo": {<geometadata>}`. One option is to store this:

```
# example_geometadata.json
{
  "columns": {},
  "primary_column": "",
  ...
}
```

scripts/update_example_schemas.py
```diff
@@ -0,0 +1,55 @@
...
+        if folder == Path("/"):
+            raise ValueError("Could not find geoparquet folder")
+
+        folder = folder.parent
+
+    return folder
+
+
+def copy_parquet_schema_metadata_to_json(parquet_file: Path) -> None:
```
Jesus89, line 24: I think it can be useful to have the opposite function too: read the JSON file and update the parquet file. This can speed up the updates while developing new versions of the spec.

Jesus89 (commented):
Thanks for the contribution!

I've added some comments. The most relevant one is the one about the content of the metadata.json file.

scripts/poetry.lock
kylebarron: On the contrary, I have a strong preference for a lockfile. Because this is a distributed project, making sure that every contributor has the exact same set of dependencies ensures that we're on the same page and not missing any small bugs.

For example I recently learned that Pyarrow v7 updated their LZ4 compression format when writing files, and so LZ4-compressed parquet files written with pyarrow v6 and pyarrow v7 are not the same.

scripts/pyproject.toml
kylebarron: I'm also in favor of standardization, though I think it's less important between the scripts and the validator if we plan for the validator to get spun off into its own repo at some point.

examples/example_metadata.json
kylebarron: My thinking is it would be better to include both, because we could have an example where we show that the user can include both `geo` metadata _and_ their own custom metadata. That answers a potential question: "I'm already using some custom metadata in my Parquet files; does that conflict with adding GeoParquet metadata?"

It's true that the metadata is stored as bytes and not as strings, but for the informative purposes of this metadata, I don't think omitting the fact that `'geo'` is of type `bytes` rather than `str` is an issue.

scripts/update_example_schemas.py
cholmes: I agree that would be cool - would definitely help me ;) 

But let's just make an issue for this and revisit in the future. I think it'd be great to get this PR for 0.2.0.w

kylebarron: +1 on making an issue and revisiting in the future

cholmes: Cool. Added #69

examples/example_metadata.json
Jesus89: OK, I think we can add later documentation clarifying the format of the JSON file, and refer to the spec for the metadata encoding.

jorisvandenbossche: Note that the `b"geo"` vs `"geo"` is only a pyarrow interface issue (which exposes metadata as bytes), the actual Parquet metadata key/values are stores as UTF8 strings, which is mentioned in the spec:

https://github.com/opengeospatial/geoparquet/blob/d8592c1d540bb9f9b3a4a4f41d6b0dc4978736e2/format-specs/geoparquet.md?plain=1#L24

(this was clarified in https://github.com/opengeospatial/geoparquet/pull/6)

jorisvandenbossche:
Another follow-up (that could be tracked in a new issue) is add some README content showing some commands how to set up the env + how to run the script?

PR #74, opened 2022-04-21
Explicit orientation field for polygon winding

felixpalmer:


felixpalmer:
Following discussion in https://github.com/opengeospatial/geoparquet/issues/46

format-specs/geoparquet.md
```diff
@@ -147,9 +148,15 @@ specify "MultiPolygon", but it is expected to specify
...
 
-#### Polygon winding
+#### orientation
 
-The winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
+This attribute indicates the winding order of polygon. Available values are:
+
+- counterclockwise: the winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
+
+If no value is set, no assertions are made about winding order, except that external rings have oppposite winding order to internal rings.
```
rouault, line 157: I would just shorthen this to
```suggestion
If no value is set, no assertions are made about winding order.
```
because there's really no guarantee that internal rings use a different winding order

Jesus89:
Remember to update:
- The validator scheme: https://github.com/opengeospatial/geoparquet/blob/main/validator/python/geoparquet_validator/schema.json
- The example (parquet, json): https://github.com/opengeospatial/geoparquet/tree/main/examples

rouault (approved):

format-specs/geoparquet.md
```diff
@@ -147,9 +148,15 @@ specify "MultiPolygon", but it is expected to specify
 ["Polygon", "MultiPolygon"]. Or if having 3D points, it is not sufficient to
 specify "Point", but it is expected to list "Point Z".
 
-#### Polygon winding
+#### orientation
```
cholmes, line 151: I think we should explain the edges=spherical case in this section a bit more. Like make it clear that we expect all spherical edges to be counterclockwise. I think that's where we ended? Even if that's not quite right I think we should re-explain what the table mentions - I believe the table should be an 'overview' of what you need to know, and the section should be a more in depth explanation, and shouldn't require you to look up at the table for all the information (like the fact that if edges=spherical then this is required)

format-specs/geoparquet.md
```diff
@@ -147,9 +148,15 @@ specify "MultiPolygon", but it is expected to specify
...
 
-The winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
+This attribute indicates the winding order of polygon. Available values are:
+
+- counterclockwise: the winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
```
cholmes, line 155: Could be nice to explain a bit more what winding order is. If there's a good web page on it then perhaps just a link to it. 

Also I feel like we should start with describing the rule, and then mention that GeoJSON does the same, but not open with 'the winding order follows geojson'.

cholmes (commented):
Added a couple little comments, but it's looking pretty good.

format-specs/geoparquet.md
brendan-ward, line 155:
```suggestion
- counterclockwise: the winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). All vertices of exterior polygon rings MUST be ordered in the counterclockwise direction and all interior rings MUST be ordered in the clockwise direction.
```

(opinion: "right-hand rule" is not helpful for interpretation; better to be explicit)

brendan-ward: > then mention that GeoJSON does the same

concur

brendan-ward (commented):
Thanks for pulling this together @felixpalmer !  Greatly appreciate your effort to move quickly on this.

I tried to add some clarifying text below.

I think the big decision point at the moment is whether orientation is conditionally required if edges="spherical".  At the moment, I'm loathe to rush that in as a hard requirement in the spec; I think most requirements should start as optional and graduate after time and vetting in the broader ecosystem.  I think in this case we can leave it optional, encourage writers to set it, and leave it up to readers to demand that it be present (required in practice rather than in specification).  That gives more time for others to weigh in that are more familiar with spherical coordinates before making it a hard requirement.

format-specs/geoparquet.md
```diff
@@ -54,9 +54,10 @@ Each geometry column in the dataset must be included in the columns field above
...
 | encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported. |
 | geometry_type | string or \[string] | **REQUIRED** The geometry type(s) of all geometries, or 'Unknown' if they are not known.  |
-| crs       | string   | **OPTIONAL** [WKT2](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html) string representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude. In the case where a crs is not provided, CRS-aware implementations should assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84) (longitude-latitude coordinates) |
+| orientation | string | ***REQUIRED** if edges='spherical', otherwise **OPTIONAL** Winding order of outer ring of polygons; interior rings are wound in opposite order. If present must be "counterclockwise". If absent, no assertions are made regarding the winding order.
```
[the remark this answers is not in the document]
cholmes: +1 on making this optional. Definitely want to add something about spherical and orientation, but may be a bit better to mention that in the edges field.

format-specs/geoparquet.md
cholmes: +1 - someone else on the thread also raised right-hand being confusing

format-specs/geoparquet.md
```diff
@@ -147,9 +148,15 @@ specify "MultiPolygon", but it is expected to specify
...
 
-#### Polygon winding
+#### orientation
 
-The winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
+This attribute indicates the winding order of polygon. Available values are:
+
+- counterclockwise: the winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
+
+If no value is set, no assertions are made about winding order.
```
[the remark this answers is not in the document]
cholmes: Great addition.

format-specs/geoparquet.md
```diff
@@ -147,9 +148,15 @@ specify "MultiPolygon", but it is expected to specify
...
+
+Writers should add the orientation attribute when it is applicable to the data.
```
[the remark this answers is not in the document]
cholmes: +1

format-specs/geoparquet.md
```diff
@@ -147,9 +148,15 @@ specify "MultiPolygon", but it is expected to specify
 ["Polygon", "MultiPolygon"]. Or if having 3D points, it is not sufficient to
 specify "Point", but it is expected to list "Point Z".
 
-#### Polygon winding
+#### orientation
 
-The winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
+This attribute indicates the winding order of polygon. Available values are:
+
+- counterclockwise: the winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). All vertices of exterior polygon rings MUST be ordered in the counterclockwise direction and all interior rings MUST be ordered in the clockwise direction.
```
jorisvandenbossche, line 155:
```suggestion
- "counterclockwise": the winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). All vertices of exterior polygon rings MUST be ordered in the counterclockwise direction and all interior rings MUST be ordered in the clockwise direction.
```

format-specs/geoparquet.md
```diff
@@ -159,6 +168,10 @@ This attribute indicates how to interpret the edges of the geometries: whether t
 
 If no value is set, the default value to assume is 'planar'.
 
+Note if edges is spherical then it is recommended that `orientation` is always set to 'counterclockwise'. If it is not set 
+then implementations should choose the smaller polygon for interpretation (but note this can only work with polygons
```
cholmes, line 172: Note I don't entirely understand this so was just trying to represent what @mentin said. If anyone (@rouault? @jorisvandenbossche?) could review / suggest changes that'd be great. But I also want to cut this release soon, and this feels like a pretty minor point, so we can also just come back and fix it in a 0.4)

cholmes (approved):
Ok, I put my latest changes in. Not sure if I represented the spherical case totally right, but hopefully either get a quick review on that or we can revisit in the next release as I think it's a minor point.

format-specs/geoparquet.md
felixpalmer: Personally, I don't really like the "choose the smaller polygon" approach. How does one compare the two polygons, with what error margin? It feels error-prone and as mentioned means we cannot have polygons larger than the hemisphere. How about for `orientation: spherical` the reader just assumes `orientation: counterclockwise`?

alasarr (dismissed):
Thanks all the people for moving this!

examples/example.py
```diff
@@ -24,14 +24,15 @@
 
 
 metadata = {
-    "version": "0.2.0",
+    "version": "0.3.0",
```
alasarr, line 27: If you're upgrading to 0.3.0 you also need to update the readme

alasarr (changes requested):

examples/example.py
alasarr: I also think we should update the PR title to say something like:
```
0.3.0 release: explicit optional orientation field for polygon winding
```

felixpalmer: Done

Jesus89 (approved):

alasarr:
I think we shouldn't mix 0.3 release with adding the orientation field as optional

format-specs/geoparquet.md
brendan-ward: I think when this field is absent, the reader can make the most pragmatic choice for the reader, and we probably shouldn't make a recommendation about what that means here.  I think from a spec standpoint, having `edges: spherical` and lacking `orientation` gives the reader permission to reject these datasets outright instead of making any assumptions, if they so choose.  Or they can make their own stated assumptions about what to do when it is absent, but that is documentation on the reader side rather than the spec.

format-specs/geoparquet.md
```diff
@@ -147,9 +148,17 @@ specify "MultiPolygon", but it is expected to specify
...
-The winding order of polygons follows the [GeoJSON spec](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.6). Polygon rings MUST follow the right-hand rule for orientation (counterclockwise external rings, clockwise internal rings).  Traversing vertices of rings in order, the interior of the polygon is on the left.
+This attribute indicates the winding order of polygons. The only available value is:
+
+- "counterclockwise": All vertices of exterior polygon rings MUST be ordered in the counterclockwise direction and all interior rings MUST be ordered in the clockwise direction.
+
+If no value is set, no assertions are made about winding order or consistency of such between exterior and interior rings or between individual geometries within a dataset.  Readers are responsible for verifying and if necessary re-ordering vertices as required for their analytical representation.
+
+Writers are encouraged but not required to set orientation="counterclockwise" for portability of the data within the broader ecosystem.
+
+It is recommended to always set the orientation (to counterclockwise) if `edges` is 'spherical'.
```
brendan-ward, line 161:
```suggestion
It is recommended to always set the orientation (to counterclockwise) if `edges` is 'spherical' (see below).
```

format-specs/geoparquet.md
```diff
@@ -159,6 +168,10 @@ This attribute indicates how to interpret the edges of the geometries: whether t
 
 If no value is set, the default value to assume is 'planar'.
 
+Note if edges is spherical then it is recommended that `orientation` is always set to 'counterclockwise'. If it is not set
```
brendan-ward, line 171:
```suggestion
Note if `edges` is 'spherical' then it is recommended that `orientation` is always set to 'counterclockwise'. If it is not set, it is not clear how polygons should be interpreted within spherical coordinate systems, which can lead to major analytical errors if interpreted incorrectly. 
```

(per my comment, suggest dropping assumed behavior)

brendan-ward (commented):
This is looking good to me overall, thanks for the continued effort here!

I think we should be careful about making any recommendation for spherical coordinates when orientation is absent, given the discussion in #46; I think we should just point out is a a good idea to have it and problematic if absent.

format-specs/geoparquet.md
felixpalmer: Sounds good to me

alasarr (approved):

examples/example.py
```diff
@@ -32,6 +32,7 @@
             "geometry_type": ["Polygon", "MultiPolygon"],
             "crs": df.crs.to_wkt(pyproj.enums.WktVersion.WKT2_2019_SIMPLIFIED),
             "edges": "planar",
+            "orientation": "counterclockwise",
```
jorisvandenbossche, line 35: This is wrong, as the example data is not actually using CCW rings. 

The current example file is converted from a Shapefile, which has by definition an opposite winding order.

jorisvandenbossche:
I opened https://github.com/opengeospatial/geoparquet/pull/80 with a short-term fix for the example file and some small clean-up (I think there was a left-over sentence from accepting an inline review comment)

PR #75, opened 2022-04-21
CI linting for consistent whitespace

kylebarron:
From https://github.com/opengeospatial/geoparquet/pull/70#discussion_r853106823

If we have consistent linting of files in the repo, ideally the git blame won't be polluted by immaterial changes from unrelated PRs.

This PR adds [pre-commit](https://pre-commit.com/) as a way to run hooks on both commit and on CI. This runs two lint checks:
- `trailing-whitespace` to assert that each line has no trailing whitespace
- `end-of-file-fixer` to assert that each file ends with a single trailing newline

To run these checks automatically on commit, you can run `pip install pre-commit && pre-commit install`.


cholmes (approved):
I'm not sure how to test this out, but I'm +1 on doing this, and it looks reasonable to me.

felixpalmer:
I wonder if it is worth having a more complete Markdown linter, like we use in deck (https://github.com/visgl/deck.gl/blob/master/.markdownlintrc). Among other things this makes things like tables look nice. Thoughts?

felixpalmer (approved):

kylebarron:
> I'm not sure how to test this out
Before this commit [`053b527` (#75)](https://github.com/opengeospatial/geoparquet/pull/75/commits/053b52726de6d8c8faa1cfa64fe64f3a52ef9a35), CI failed https://github.com/opengeospatial/geoparquet/runs/6115837709?check_suite_focus=true, and after that commit CI passed https://github.com/opengeospatial/geoparquet/runs/6115849790?check_suite_focus=true. So that shows that the linting is working on CI.

kylebarron:
> I wonder if it is worth having a more complete Markdown linter, like we use in deck

I'm always +1 on linters and formatters. The actual choice of formatting is less important than the consistency it provides.

Not sure if a markdown linter would remove the need for checking trailing whitespace, since this also checks trailing whitespace in non-markdown files.

jorisvandenbossche (approved):

kylebarron:
@cholmes If you install `pre-commit`, with `pip install pre-commit`. Then you can run `pre-commit install` once to automatically run these checks whenever you make a git commit. Or to run them manually you can run `pre-commit run --all-files`

cholmes:
Cool, thanks - sorry I was messing things up. The way to manually run was a mystery to me - we should probably document this somewhere. I can try to take it on.

cholmes:
When the test 'fails' does it still change the files? I see:

```
Trim Trailing Whitespace.................................................Failed
- hook id: trailing-whitespace
- exit code: 1
- files were modified by this hook

Fixing README.md
```

At that point do I just run the commit again? Like it will have linted my files?

kylebarron:
When the test fails it fixes the files but leaves the changes _unstaged_. So you just need to stage the file again with `git add README.md`.

cholmes:
Awesome, thanks!

PR #76, opened 2022-04-21
Add README for scripts folder

kylebarron:
For https://github.com/opengeospatial/geoparquet/pull/68#issuecomment-1103178777

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -1,0 +1,17 @@
+# Helper scripts
+
+## Usage
+
+The scripts in this directory use [Poetry](https://github.com/python-poetry/poetry) for describing dependencies and keeping a consistent lockfile. This lockfile is useful because it ensures every contributor is able to use the exact same dependencies.
+
+To install Poetry, follow the Poetry [installation guide](https://python-poetry.org/docs/#installation).
+
+To install from the lockfile, run `poetry update`.
+
+To run a script, prefix it with `poetry run`. For example:
+
+```
+poetry run python update_example_schemas.py
+```
+
+Using `poetry run` ensures that you're running the python script using _this_ local environment, not your global environment.


cholmes (approved):
Ran through the instructions and it worked great. Thanks!

felixpalmer (approved):
Tried it and works well for me also

scripts/README.md
```diff
@@ -0,0 +1,17 @@
+# Helper scripts
+
+## Usage
+
+The scripts in this directory use [Poetry](https://github.com/python-poetry/poetry) for describing dependencies and keeping a consistent lockfile. This lockfile is useful because it ensures every contributor is able to use the exact same dependencies.
+
+To install Poetry, follow the Poetry [installation guide](https://python-poetry.org/docs/#installation).
+
+To install from the lockfile, run `poetry update`.
```
jorisvandenbossche, line 9: Is this `poetry update` step supposed to create a new environment? 

Because that doesn't seem to be what it is doing, in the way I tried it. We have a conda environment.yml file in the examples/ directory to create the example.parquet file. And so that environment was activated when, after updating the example, I ran those instructions. And it seems to have updated my conda installation instead of creating a separate virtual environment.

jorisvandenbossche (commented):
@kylebarron thanks for adding some instructions!

scripts/README.md
kylebarron: 🤦‍♂️ Sorry, I wrote this doc too quickly. To install from a lockfile you use `poetry install`, not `poetry update`. The latter updates the lockfile with new dependencies (but still constrained to the semver dependencies in `pyproject.toml`).

I don't use Conda so I'm not familiar of any ways Poetry could interact with it. One way to check how Poetry is behaving is by setting the config settings to create a virtualenv in the project directory:
```
poetry config virtualenvs.create true
poetry config virtualenvs.in-project true
```

Then when I run `poetry install` within `scripts/` I see this:
```
> poetry install
Creating virtualenv geoparquet-scripts in /Users/kyle/github/mapping/geoparquet/scripts/.venv
Installing dependencies from lock file

Package operations: 2 installs, 0 updates, 0 removals

  • Installing numpy (1.22.3)
  • Installing pyarrow (7.0.0)
```
And then you should see a `.venv` directory in the local directory, and be able to open that environment (even outside of `poetry run`) with a normal `source ./.venv/bin/activate`.

kylebarron: That said, I suggested this setup because it's what I'm most familiar with, but it's ok if the project standardizes on another system. Mostly I like having the reproducibility of lockfiles while only needing to strictly define top-level dependencies (see: [ref1](https://modelpredict.com/wht-requirements-txt-is-not-enough), [ref2](https://nvie.com/posts/better-package-management/)). Poetry, Pipenv, and pip-compile+requirements.txt+virtualenv all have similar goals.

We could explore [conda-lock](https://github.com/conda-incubator/conda-lock).

PR #77, opened 2022-04-22
updates for 0.3 release

cholmes:
Updates to readme, examples, validators, etc. for 0.3.0 release.


felixpalmer (approved):

README.md
```diff
@@ -60,8 +60,8 @@ will work much better if it is backing a system that is constantly updating the
 Our aim is to get to a 1.0.0 within 'months', not years. The rough plan is:
 
 * 0.1 - Get the basics established, provide a target for implementations to start building against.
-* 0.2 - Feedback from implementations, 3D coordinates support, geometry types, crs optional.
-* 0.3 - Feedback from implementations, add spatial index.
+* 0.2 / 0.3 - Feedback from implementations, 3D coordinates support, geometry types, crs optional.
```
kylebarron, line 63: I think it would be great to add a Changelog file, which describes at a glance exactly what is in each version, and highlights what are the backwards incompatible changes in each release. For example we can say for `0.2` -> `0.3`, "we received feedback from the community that enforcing winding order was too strict, and wanted to make a quick new release to make that optional".

I think having this file would be beneficial for people who aren't closely following all our discussions.

kylebarron (approved):

alasarr (changes requested):
We need to update too:
- `geoparquet.md`. https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#version
- `example_metadata.json`
- `example.py`
-  example parquet file

README.md
cholmes: Note that the release notes have a pretty nicely auto-generated changelog - https://github.com/opengeospatial/geoparquet/releases/tag/v0.2.0

In stac-spec we did maintain a changelog by hand, so I was very pleased to see with the auto-generated one. Perhaps there's some way to just raise the visibility of that more?

alasarr (approved):

README.md
kylebarron: We could potentially copy that into a `CHANGELOG.md` file 🤷‍♂️

kylebarron: Also note that the auto-generated changelog is somewhat distinct from what deck.gl calls an "upgrade guide", which describes _only_ the breaking changes needed to upgrade existing apps: https://deck.gl/docs/upgrade-guide

PR #85, opened 2022-04-25
Added a changelog

cholmes:
From suggestion in https://github.com/opengeospatial/geoparquet/pull/77#discussion_r856388158 I made an attempt at just copying and pasting a changelog. 

It seems like it might be useful. Making this PR to discuss, to see if it's something we want - no problem if we decide not to, just wanted to see how it would actually work.

I was thinking we could just use GitHub's release note autogenerator to create it, but it looks like you need to add the tag first. So if we wanted the changelog to appear on a release I think we'd need to partially cut a release by making the tag, then make a PR with the changes to the changelog generated from that tag, and then I guess update the tag to include the changelog PR. Then fully cut the release.

So definitely more annoyance on release time, but probably better than making people add each change by hand when they make a PR. Or we could do that and adopt the [keep a changelog](https://keepachangelog.com/en/1.0.0/) form and get the added / changed / etc sections right - this doesn't do that since it's fully autogenerated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,55 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
+It is mostly auto-generated from GitHub's release notes.
+
+# [v0.2.0] - 2022-04-19
+
+## What's Changed
+* Add Apache license by @cholmes in https://github.com/opengeospatial/geoparquet/pull/38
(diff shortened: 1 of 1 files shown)


cholmes:
Hrm, sorry about the lint failing - I thought I installed the pre-commit check, but it didn't seem to work.

cholmes:
I'm going to close this, I didn't love the results, we can revisit later.

Issue #42, opened 2022-03-10
nz-buildings-outlines.parquet sample file uses 'schema_version' instead of 'version'


rouault:
https://storage.googleapis.com/open-geodata/linz-examples/nz-buildings-outlines.parquet has the following 'geo' metadata value:
```json
{
  "primary_column": "geometry",
  "columns": {
    "geometry": {
      "crs": "PROJCRS[\"NZGD2000 / New Zealand Transverse Mercator 2000\",BASEGEOGCRS[\"NZGD2000\",DATUM[\"New Zealand Geodetic Datum 2000\",ELLIPSOID[\"GRS 1980\",6378137,298.257222101,LENGTHUNIT[\"metre\",1]]],PRIMEM[\"Greenwich\",0,ANGLEUNIT[\"degree\",0.0174532925199433]],ID[\"EPSG\",4167]],CONVERSION[\"New Zealand Transverse Mercator 2000\",METHOD[\"Transverse Mercator\",ID[\"EPSG\",9807]],PARAMETER[\"Latitude of natural origin\",0,ANGLEUNIT[\"degree\",0.0174532925199433],ID[\"EPSG\",8801]],PARAMETER[\"Longitude of natural origin\",173,ANGLEUNIT[\"degree\",0.0174532925199433],ID[\"EPSG\",8802]],PARAMETER[\"Scale factor at natural origin\",0.9996,SCALEUNIT[\"unity\",1],ID[\"EPSG\",8805]],PARAMETER[\"False easting\",1600000,LENGTHUNIT[\"metre\",1],ID[\"EPSG\",8806]],PARAMETER[\"False northing\",10000000,LENGTHUNIT[\"metre\",1],ID[\"EPSG\",8807]]],CS[Cartesian,2],AXIS[\"northing (N)\",north,ORDER[1],LENGTHUNIT[\"metre\",1]],AXIS[\"easting (E)\",east,ORDER[2],LENGTHUNIT[\"metre\",1]],USAGE[SCOPE[\"Engineering survey, topographic mapping.\"],AREA[\"New Zealand - North Island, South Island, Stewart Island - onshore.\"],BBOX[-47.33,166.37,-34.1,178.63]],ID[\"EPSG\",2193]]",
      "encoding": "WKB",
      "bbox": [
        1167512.311218509,
        4794679.949937864,
        2089113.650566361,
        6190596.90070761
      ]
    }
  },
  "schema_version": "0.1.0",
  "creator": {
[...]

PR #87, opened 2022-04-27
Script to write nz-building-outlines to geoparquet 0.4.0

kylebarron:
@cholmes asked me to write a script to update the `nz-building-outlines` file to Parquet version 0.4.0. This reads from the GeoPackage version of the data, which you can download from [here](https://storage.googleapis.com/open-geodata/linz-examples/nz-building-outlines.gpkg) (1.3GB). 

Part of this is derived from the GeoPandas code [here](https://github.com/geopandas/geopandas/blob/main/geopandas/io/arrow.py). But this additionally ~~uses [pyogrio](https://github.com/geopandas/pyogrio)~~ (reverted because it failed to build on CI) and pygeos to try and speed things up a little bit. It's probably not a bad thing to have a Python script here because the GeoPandas release schedule is likely slower than our release schedule (at least thus far).

You can install the new dependencies with `poetry install` and run it with

```bash
poetry run python write_nz_building_outline.py \
    --input nz-building-outlines.gpkg \
    --output nz-building-outlines.parquet \
    --compression SNAPPY
```
This takes about 5 minutes on my computer. With Snappy compression, the 1.3GB GeoPackage file became 410MB in Parquet. (375MB with ZSTD compression).

To see the CLI options you can run

```bash
poetry run python write_nz_building_outline.py --help
```

Closes https://github.com/opengeospatial/geoparquet/issues/42


jorisvandenbossche:
Thanks for working on this! It's indeed good to have a script here to create files, that doesn't depend on geopandas' `to_parquet`, to be able to be more up to date.

One question: do we want to make this a "generic" script for converting something to GeoParquet? Or specific to this file? Because now it is partly specific (the file name, the dtype casting), but also has some generic code from geopandas that could be simplified if it's only targetting this file.

> But this additionally uses [pyogrio](https://github.com/geopandas/pyogrio) (reverted because it failed to build on CI)

Yeah, as long as we don't have wheel (hopefully soon!), you need to have GDAL installed to built it from source (or easier, use conda ;))

kylebarron:
> One question: do we want to make this a "generic" script for converting something to GeoParquet? Or specific to this file? Because now it is partly specific (the file name, the dtype casting), but also has some generic code from geopandas that could be simplified if it's only targetting this file.

Yeah I figured I'd make the PR and then let people discuss 🙂 . As you mentioned most of the script is generalized except for dtype casting.

cholmes:
This is great! Will try it out when I get a chance, hopefully soon.

I agree it'd be great to have a generic script, as long as it had a bit of documentation for how to use it to create the nz-buildings one. The dtype casting does look a bit complicated, so curious how that could be done generically? Some optional config file? Some command-line options?

kylebarron:
> The dtype casting does look a bit complicated, so curious how that could be done generically? Some optional config file? Some command-line options?

You usually need dtype casting when loading data from csv, because csv is an untyped data format (everything is stored as string). In this case, the data as loaded from the GeoPackage file didn't match linz's [public description of the data](https://nz-buildings.readthedocs.io/en/latest/published_data.html#table-nz-building-outlines). Specifically a couple columns where int64 instead of int32, and three columns had been stored as strings instead of dates. For a file that we want to present as an example of good Parquet usage, I think it would be good to distribute it with as efficient dtypes as possible.

cholmes:
Ok, just tried this out. It seemed to work? But there were a couple little things:

```
(base) cholmes@x9t37m3hgt scripts % poetry run python write_nz_building_outline.py \
    -i nz-building-outlines.gpkg \
    -o nz-building-outlines.parquet \
    --compression SNAPPY
/Users/cholmes/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-DKHrdpBy-py3.9/lib/python3.9/site-packages/geopandas/_compat.py:111: UserWarning: The Shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the GEOS version PyGEOS was compiled with (3.10.1-CAPI-1.16.0). Conversions between both will be slow.
  warnings.warn(
Starting to read geopackage
Finished reading geopackage
Starting conversion to Arrow
/Users/cholmes/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-DKHrdpBy-py3.9/lib/python3.9/site-packages/pygeos/io.py:85: UserWarning: The shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the PyGEOS GEOS version (3.10.1-CAPI-1.16.0). Conversions between both will be slow
  warnings.warn(
/Users/cholmes/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-DKHrdpBy-py3.9/lib/python3.9/site-packages/geopandas/geodataframe.py:1350: UserWarning: Geometry column does not contain geometry.
  warnings.warn("Geometry column does not contain geometry.")
Finished conversion to Arrow
Starting write to Parquet
Finished write to Parquet
```

It did seem to take awhile, like longer than your 3 minutes. It warned about that with different GEOS installs. Though not sure how I got that / how I fix it - this machine is only a month or two old.

And then I'm not sure about the 'geometry column does not contain a geometry', if something went wrong? Is there a way to verify the output? It came out as 410MB, so in line with yours.

kylebarron:
Ok, just tried this out. It seemed to work? But there were a couple little things:

(base) cholmes@x9t37m3hgt scripts % poetry run python write_nz_building_outline.py \
    -i nz-building-outlines.gpkg \
    -o nz-building-outlines.parquet \
    --compression SNAPPY
/Users/cholmes/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-DKHrdpBy-py3.9/lib/python3.9/site-packages/geopandas/_compat.py:111: UserWarning: The Shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the GEOS version PyGEOS was compiled with (3.10.1-CAPI-1.16.0). Conversions between both will be slow.
  warnings.warn(
Starting to read geopackage
Finished reading geopackage
Starting conversion to Arrow
/Users/cholmes/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-DKHrdpBy-py3.9/lib/python3.9/site-packages/pygeos/io.py:85: UserWarning: The shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the PyGEOS GEOS version (3.10.1-CAPI-1.16.0). Conversions between both will be slow
  warnings.warn(
/Users/cholmes/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-DKHrdpBy-py3.9/lib/python3.9/site-packages/geopandas/geodataframe.py:1350: UserWarning: Geometry column does not contain geometry.
  warnings.warn("Geometry column does not contain geometry.")
Finished conversion to Arrow
Starting write to Parquet
Finished write to Parquet

> It did seem to take awhile, like longer than your 3 minutes.

Yeah my 3 minutes was when I initially used `pyogrio`. But I reverted that because it caused CI to fail. So now reading the GeoPackage might take an extra couple minutes.

> It warned about that with different GEOS installs.

This happens between Shapely and PyGEOS because they each have different versions of GEOS that they build against. It's not an issue with your machine. I did try to add a note to the README of how to recompile pygeos, which should fix the issue:

https://github.com/kylebarron/geoparquet/blob/77a0953728af8919f34ed63d198aaed34173cc5f/scripts/README.md?plain=1#L38-L46

> And then I'm not sure about the 'geometry column does not contain a geometry'

No, that's expected. I believe that happens here:
https://github.com/kylebarron/geoparquet/blob/77a0953728af8919f34ed63d198aaed34173cc5f/scripts/write_nz_building_outline.py#L151-L152
Because I'm reassigning the GeoPandas geometry column to now be a binary buffer (the WKB data) instead of GeoPandas' native geometry types. I'm sure there's a way to do this properly such that it doesn't emit a warning.

> Is there a way to verify the output?

Hmm, I'm not sure. You can verify the metadata against the json schema at least. You could use `geopandas.read_parquet()` to make sure the data could be read back into a `GeoDataFrame`.

cholmes:
Ok, great. As long as the geometry column thing is expected I'll just go ahead and put this one online after verifying the metadata.

kylebarron:
I'd love for someone to check my work 🙂. We can probably defer the question of "generic script vs nz-building-outlines -specific" for later. If desired, it wouldn't be hard to make into a small installable Python library to do `GeoDataFrame -> GeoParquet`. Not sure if there's an appetite for that outside of the GeoPandas library though.

Jesus89:
I have tested on my machine the generation of the parquet file (411MB).

```
$ time poetry run python write_nz_building_outline.py     --input nz-building-outlines.gpkg     --output nz-building-outlines.parquet     --compression SNAPPY
/home/jesus/code/geoparquet/validator/python/venvt/lib/python3.8/site-packages/geopandas/_compat.py:111: UserWarning: The Shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the GEOS version PyGEOS was compiled with (3.10.1-CAPI-1.16.0). Conversions between both will be slow.
  warnings.warn(
Starting to read geopackage
Finished reading geopackage
Starting conversion to Arrow
/home/jesus/code/geoparquet/validator/python/venvt/lib/python3.8/site-packages/pygeos/io.py:85: UserWarning: The shapely GEOS version (3.10.2-CAPI-1.16.0) is incompatible with the PyGEOS GEOS version (3.10.1-CAPI-1.16.0). Conversions between both will be slow
  warnings.warn(
/home/jesus/code/geoparquet/validator/python/venvt/lib/python3.8/site-packages/geopandas/geodataframe.py:1350: UserWarning: Geometry column does not contain geometry.
  warnings.warn("Geometry column does not contain geometry.")
Finished conversion to Arrow
Starting write to Parquet
Finished write to Parquet

real    6m7,805s
user    6m2,748s
sys     0m5,148s
```

IMO it solves the original task, which is having a script to update `nz-building-outlines`. I agree with talking about a generic script later.

---

I noticed that `poetry install` removes `colorama 0.4.4` which is a dependency of `geoparquet_validator`. Could you include temporarily `colorama = "^0.4.4"` in the `pyproject.toml` file, or provide another solution to make this compatible with the validator?

cholmes (approved):

kylebarron:
> I noticed that `poetry install` removes `colorama 0.4.4` which is a dependency of `geoparquet_validator`. Could you include temporarily `colorama = "^0.4.4"` in the `pyproject.toml` file, or provide another solution to make this compatible with the validator?

I'm not sure I follow. Currently the Poetry setup is completely separate from the validator, and it's expected for these to run in two different virtual environments. The validator is not installed within this pyproject.toml, so it doesn't know about any of its dependencies.

Jesus89:
I was using the same virtual environment for the validator and for this script, just to validate the file I generated. But if the idea is to use different virtual environments, then there is no issue.

kylebarron:
Yeah as-is it's expected to use two different environments. If we want to use the same environment, we can declare the validator a dependency of this `pyproject.toml.

kylebarron:
> But there were a couple little things

I updated to hide the "column does not contain geometry" warning. Or maybe it would be better instead of ignoring the warning to construct a pandas DataFrame instead of modifying the GeoDataFrame 😄 

I suppose now that GDAL is about to ship with a GeoParquet driver, converting other files to GeoParquet should be easier, but that might not always be up to date with the latest version of the spec.

cholmes:
Can we merge this? Will it need any updates for 0.4.0? Or will it just work once geopandas is updated?

kylebarron:
I think the main question outstanding is how generic we want this to be. Probably fine to merge as it is now and follow up in the future. Would love another ✅ review though.

> will it just work once geopandas is updated?

This doesn't rely on geopandas for actually writing the Parquet file. Using Geopandas was just easier to load the GeoPackage into memory. So it shouldn't require an updated version of geopandas.

> Will it need any updates for 0.4.0?

I assume Projjson will be in 0.4.0, and then this script will need to change to store that instead of the WKT in the metadata.

cholmes:
> I think the main question outstanding is how generic we want this to be. Probably fine to merge as it is now and follow up in the future. 

Yeah, I think we should merge as is, and then in a future version make it more generic.

> Would love another ✅ review though.

Agreed - someone who actually writes python should give a thumbs up.

> So it shouldn't require an updated version of geopandas.

Great

>> Will it need any updates for 0.4.0?

> I assume Projjson will be in 0.4.0, and then this script will need to change to store that instead of the WKT in the metadata.

Yeah, prjjson will be in 0.4.0. Would be great if you could update this PR to include that.

scripts/write_nz_building_outline.py
```diff
@@ -0,0 +1,209 @@
...
+
+def parse_to_pygeos(df: gpd.GeoDataFrame) -> Dict[str, PygeosGeometryArray]:
+    """Parse to pygeos geometry array
+
+    This is split out from _create_metadata so that we don't have to create the pygeos
+    array twice: once for converting to wkb and another time for metadata handling.
+    """
+    geometry_columns: Dict[str, PygeosGeometryArray] = {}
+    for col in df.columns[df.dtypes == "geometry"]:
+        pygeos_geoms = pygeos.from_shapely(df[col])
```
jorisvandenbossche, line 54: This step is not really needed, because if pygeos is installed, geopandas will automatically use pygeos under the hood (so this step is basically converting the pygeos geometries to shapely (because `pygeos.from_shapely` converts the Series to a numpy object array, which will output shapely geometries) and then back to pygeos geometries)

jorisvandenbossche: You can get the numpy array with pygeos geometries directly as `df[col].array.data`

scripts/write_nz_building_outline.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def geopandas_to_arrow(df: gpd.GeoDataFrame) -> pa.Table:
+    geometry_columns = parse_to_pygeos(df)
+    geo_metadata = _create_metadata(df, geometry_columns)
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message=".*does not contain geometry.*")
+        for col, geometry_array in geometry_columns.items():
+            df[col] = pygeos.to_wkb(geometry_array)
```
jorisvandenbossche, line 155:
```suggestion
    df = pd.DataFrame(df)
    for col, geometry_array in geometry_columns.items():
        df[col] = pygeos.to_wkb(geometry_array)
```

Alternative to avoid the warning

jorisvandenbossche (commented):
Some code comments. But it's also fine to handle those as a follow-up, if this is already used to create the file we upload, then it should be fine to already merge this (and do the updates for 0.4.0 in a separate PR?)

scripts/write_nz_building_outline.py
kylebarron: Thanks!

kylebarron:
It would probably be good to test this output against the v0.4 json schema validator as well

kylebarron:
I was able to verify that the output of this validated against the current validator

cholmes:
Awesome. I've been hit with a morning of meetings, but I'm +1 on merging this, with a final thumbs up from @jorisvandenbossche or another python dev. I'll aim to try it out when I get some time between meetings.

Jesus89 (approved):
Hey, I've tested this again, with the latest changes. I've been able to generate `nz-building-outlines.parquet` and it works properly, also using the validator to check the file.

I would suggest adding the link of the input source in the README file to make this reproducible, but it can be added in another pull request to move this forward (https://storage.googleapis.com/open-geodata/linz-examples/nz-building-outlines.gpkg).

**Note**

Moving the schema to `format-specs` broke the validator: `pip install .` does not work with a symlink resource. It works with `pip install -e .`, but this is something we need to tackle, especially if we want to distribute the validator as a python package.

cholmes:
> Note
Moving the schema to format-specs broke the validator: pip install . does not work with a symlink resource. It works with pip install -e ., but this is something we need to tackle, especially if we want to distribute the validator as a python package.

Shoot, I somehow missed reading this before cutting the 0.4.0 release. I suppose it's not strictly necessary. But would be something that'd be good to fix soon. Made an issue #107 to track it.

PR #94, opened 2022-05-09
Allow the "crs" to be unknown ("crs": null)

jorisvandenbossche:
From https://github.com/opengeospatial/geoparquet/issues/52#issuecomment-1117463839

Issue #26, opened 2022-03-03
Use RFC 2119 definitions of MUST, SHOULD, etc.?


cholmes:
A number of specs specifically say something like: 

The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) See [mbtiles spec](https://github.com/mapbox/mbtiles-spec/blob/master/1.3/spec.md#abstract). for example.

Should we do the same? And go through our language to make sure it matches? I'm not sure that we ever did this for STAC. But I think OGC specs usually do.

I don't feel strongly, but making an issue so we discuss and track.

m-mohr:
STAC tries to use this, but did never fully embrace it due to the amount of text.
Here it could be feasible as the spec is much more lightweight, so +1.

PR #96, opened 2022-05-11
Use PROJJSON instead of WKT2:2019

### LINKED ISSUE #95: Use projjson instead of wkt2
Lots of discussion on this issue in https://github.com/opengeospatial/geoparquet/discussions/90 and also reached full consensus at meeting 5/10/22, notes at https://docs.google.com/document/d/19J2wsEdjyOP5pIOATWpZIKnHcbxFFAnxQtY49PlS_OM/edit#

This ticket is to track actually including it in the spec.

brendan-ward:
Resolves #95 based on associated discussion in #90.

This assumes #94 is added separately, allowing `crs` to be explicitly set to null.

This preserves the existing optional behavior and default if unspecified of `OGC:CRS84`.  The default value and optional vs required status will be taken up in another PR.

This includes a minor formatting change of the table including `crs` but only the `crs` row changed content.


format-specs/geoparquet.md
```diff
@@ -50,55 +50,133 @@ Version of the geoparquet spec used, currently 0.3.0
...
+| geometry_type | string or \[string] | **REQUIRED** The geometry type(s) of all geometries, or 'Unknown' if they are not known.                                                                                                                                                                                                                                                                                                                                                                     |
+| crs           | JSON object         | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) JSON object representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude, and CRS-aware implementations may assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84) (longitude-latitude coordinates). |
```
brendan-ward, line 57: Note change with respect to default value from "should" to "may", but the specific term should likely follow [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) with respect to these terms as per #26 

Based on discussions around this default, it may instead be appropriate to use "must" in order to prevent ambiguity regarding an unset value.

kylebarron:
Pending #93 (and #64, a discussion of whether to make the json schema definition a core part of the geoparquet metadata spec), we should decide whether we want the json schema to be updated _along with_ or _after_ associated spec PRs. In this case, we'd want the json schema to be updated so that the `crs` refers to the projjson json schema: https://proj.org/schemas/v0.4/projjson.schema.json

format-specs/geoparquet.md
```diff
@@ -50,55 +50,133 @@ Version of the geoparquet spec used, currently 0.3.0
...
 
 If CRS is not provided, then all coordinates in the geometry must use longitude, latitude to store their data.
-If an implementation is CRS-aware and needs a CRS representation of the data it should assume a default value is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). It's equivalent to the well-known [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) but changes the axis from latitude-longitude to longitude-latitude. The WKT2:2019 string for OGC:CRS84 is:
+If an implementation is CRS-aware and needs a CRS representation of the data it should assume a default value is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which is equivalent to the well-known [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) but changes the axis from latitude-longitude to longitude-latitude.
+
+Note: EPSG:4326 and OGC:CRS84 are equivalent with respect to this specification because this specification specifically overrides the coordinate axis order in the `crs` to be longitude-latitude.
+
+The current PROJJSON JSON object for OGC:CRS84 is:
 
 ```
```
kylebarron, line 76: nit: apply JSON syntax highlighting to the block
```suggestion
```json
```

kylebarron (approved):

brendan-ward:
> we should decide whether we want the json schema to be updated along with or after associated spec PRs

I can update the schema here once #93 is merged; I was thinking that ideally both the spec and required changes to the schema would get updated within the same PR.  It may be good for us to come up with a checklist of required / recommended updates to files here as part of making a PR to update the spec.

I'm assuming that we'd want to update the script for building an example parquet file _after_ such PRs are in place, since that corresponds more directly to a specific version of the spec rather than intermediate changes within a PR, right?

kylebarron:
> It may be good for us to come up with a checklist of required / recommended updates to files here as part of making a PR to update the spec

👍 would be a great thing for a PR template.

> I'm assuming that we'd want to update the script for building an example parquet file after such PRs are in place, since that corresponds more directly to a specific version of the spec rather than intermediate changes within a PR, right?

That's true, there are really three parts:

- Spec description in markdown
- Metadata description in JSON Schema
- Script to build parquet file

I don't have a strong opinion on when to update the Parquet script, but that could also be related to https://github.com/opengeospatial/geoparquet/discussions/40. For example, I think STAC spec developed entirely on a `dev` branch and only merged that into the main branch on each release? So when a member of the public visited the repo, they'd see the state of the last release by default instead of the current WIP state between releases.

format-specs/geoparquet.md
```diff
@@ -50,57 +50,160 @@ Version of the geoparquet spec used, currently 0.3.0
...
 
 If CRS is not provided, then all coordinates in the geometry must use longitude, latitude to store their data.
-If an implementation is CRS-aware and needs a CRS representation of the data it should assume a default value is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). It's equivalent to the well-known [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) but changes the axis from latitude-longitude to longitude-latitude. The WKT2:2019 string for OGC:CRS84 is:
 
+If an implementation is CRS-aware and needs a CRS representation of the data it may assume a default value is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which is equivalent to the well-known [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) but changes the axis from latitude-longitude to longitude-latitude.  However, there is no guarantee that coordinates
```
brendan-ward, line 71: Note: The "However..." can be removed if we instead state above that 

```
If CRS is not provided, then all coordinates in the geometry must use longitude, latitude based on the OGC:CRS84 CRS to store their data.
```

But without that, lacking a `crs` means that coordinates are in an unspecified CRS.

format-specs/geoparquet.md
```diff
@@ -50,57 +50,160 @@ Version of the geoparquet spec used, currently 0.3.0
...
+fully parse PROJJSON, it may be possible to infer that coordinates conform to
+the OGC:CRS84 CRS based on elements of the `crs` field.  For simplicity, Javascript
+object dot notation is used to refer to nested elements.
+
+The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` element is present:
+
+* `id.authority` = `"OGC"` and `id.code` = `"CRS84"`
+* `id.authority` = `"EPSG"` and `id.code` = `4326` (due to longitude, latitude ordering in this specification)
+
+The CRS is likely equivalent to OGR:CRS84 if all of the following are true:
```
brendan-ward, line 197: @rouault I'd appreciate your thoughts on the following rules for at least the common cases (we don't need exhaustive rules here).  In particular, I don't know if the units or directions are likely to always be present or even necessary (they are missing from the OGC:CRS84 xml linked above).

format-specs/geoparquet.md
```diff
@@ -50,55 +50,133 @@ Version of the geoparquet spec used, currently 0.3.0
...
+    "east_longitude": 180
+  },
+  "id": {
+    "authority": "OGC",
+    "code": "CRS84"
+  }
+}
 ```
 
+Note: earlier versions of OGC:CRS84 used a single datum instead of a datum ensemble.  The number of elements in the datum ensemble may vary over time when new realizations of WGS 84 are added to the ensemble
```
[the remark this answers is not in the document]
rouault: OGC:CRS84 official has probably frozen for > 10 years. In PROJ, we define it exactly as EPSG:4326 with just a change in the axis order, hence the definition isn't fully identical with the original OGC one and will evolve as EPSG:4326. Which of the ancient "official" / frozen OGC:CRS84 definition or the one returned by PROJ is correct is left for debate to geodesists which will ultimately disagree between themselves :-)

format-specs/geoparquet.md
```diff
@@ -50,57 +50,160 @@ Version of the geoparquet spec used, currently 0.3.0
...
+* `id.authority` = `"EPSG"` and `id.code` = `4326` (due to longitude, latitude ordering in this specification)
+
+The CRS is likely equivalent to OGR:CRS84 if all of the following are true:
+* `"type"` = `"GeographicCRS"`
+* `coordinate_system.axis[0].unit` = `"degree"`
+* `coordinate_system.axis[1].unit` = `"degree"`
+* the values for `coordinate_system.axis[n].direction` are `"east"` and `"north"` (in either order, for n in [0,1])
+
+and at least one of the following are true:
+* `datum.id.authority` = `"EPSG"` and `datum.id.code` = `6326`
```
rouault, line 204:
```suggestion
* `datum.id.authority` = `"EPSG"` and `datum.id.code` = `6326`
* `datum_ensemble.id.authority` = `"EPSG"` and `datum_ensemble.id.code` = `6326`
```

format-specs/geoparquet.md
```diff
@@ -50,57 +50,160 @@ Version of the geoparquet spec used, currently 0.3.0
...
+* `datum_enemble.ellipsoid.semi_major_axis` = `6378137` and `datum_ensemble.ellipsoid.inverse_flattening` = `298.257223563`
```
rouault, line 205:
```suggestion
* `datum_ensemble.ellipsoid.semi_major_axis` = `6378137` and `datum_ensemble.ellipsoid.inverse_flattening` = `298.257223563`
```

brendan-ward:
> OGC:CRS84 official has probably frozen for > 10 years

Which form should we include here?  The `EPSG:4326` with lon,lat ordering as we have here as a proxy for what OGC:CRS84 would be if it wasn't frozen, since that is what users get if they use PROJ to get the PROJJSON, or the form that would represent the single-datum version?  Or rather, were you just suggesting that we rephrase the note?

I think one of our goals in including it here is that at some point we can also make a recommendation for non-CRS aware tools that can safely assume they are producing OGC:CRS84 referenced data to copy / paste this json to their `crs` field (if we require `crs` or at least strongly recommend it in order to make geographic coordinates less ambiguous than now).

rouault:
For non-CRS aware tools, perhaps we should recommend the simplest minimal reasonable form of OGC:CRS84 definition
```json
{
    "type": "GeographicCRS",
    "name": "WGS 84 longitude-latitude",
    "datum": {
        "type": "GeodeticReferenceFrame",
        "name": "World Geodetic System 1984",
        "ellipsoid": {
            "name": "WGS 84",
            "semi_major_axis": 6378137,
            "inverse_flattening": 298.257223563
        }
    },
    "coordinate_system": {
        "subtype": "ellipsoidal",
        "axis": [
        {
            "name": "Geodetic longitude",
            "abbreviation": "Lon",
            "direction": "east",
            "unit": "degree"
        },
        {
            "name": "Geodetic latitude",
            "abbreviation": "Lat",
            "direction": "north",
            "unit": "degree"
        }
        ]
    },
    "id": {
        "authority": "OGC",
        "code": "CRS84"
    }
}
```

format-specs/geoparquet.md
```diff
@@ -50,57 +50,94 @@ Version of the geoparquet spec used, currently 0.3.0
...
+| epoch         | double              | **OPTIONAL** Coordinate epoch in case of a dynamic CRS, expressed as a decimal year.
 
 #### crs
 
 The Coordinate Reference System (CRS) is an optional parameter for each geometry column defined in geoparquet format.
 
-The CRS must be provided in [WKT](https://en.wikipedia.org/wiki/Well-known_text_representation_of_coordinate_reference_systems) version 2, also known as **WKT2**. WKT2 has several revisions, this specification only supports [WKT2_2019](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html).
+The CRS must be provided in [PROJJSON](https://proj.org/specifications/projjson.html) format, which is a JSON encoding of [WKT2:2019 / ISO-19162:2019](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html), which itself implements the model of [OGC Topic 2: Referencing by coordinates abstract specification / ISO-19111:2019](http://docs.opengeospatial.org/as/18-005r4/18-005r4.html).  Apart from the difference of encodings, the semantics is intended to be exactly the same as WKT2:2019, and PROJJSON can be morphed losslessly from/into WKT2:2019.
```
cholmes, line 67: Note there was a comment on this in #97 - https://github.com/opengeospatial/geoparquet/pull/97#discussion_r873977007  That's the 'require' PR, but this one should merge first so we can cut 0.4.0 release.

brendan-ward: Either one could theoretically go out as 0.4.0; per the discussion I thought the intent was to compare the optional PROJJSON with OGC:CRS84 default vs required PROJJSON, rather than them being additive.  But I can certainly update this PR to address comments in #97 if we'd like to see this particular PR go out sooner.

cholmes: Sounds good. I think the plan was to aim for 0.4.0 without it - I organized [the 0.4 milestone](https://github.com/opengeospatial/geoparquet/milestone/6) around that - and then to be sure to solve this for 0.5. Just to be sure to make iterative progress as it seemed like that one could take longer. 

I think we targeted next week for the 0.4.0 release. I'm also not opposed to get this into 0.4 release with everyone happy, and just put up a discussion in #98 to help push it forward. So maybe see where that discussion is at on monday of next week, and if it's not fully resolved then it'd be great if you could clean up this PR and we push out 0.4

format-specs/geoparquet.md
```diff
@@ -50,57 +50,100 @@ Version of the geoparquet spec used, currently 0.3.0
...
+| ------------- | ------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| encoding      | string              | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported.                                                                                                                                                                                                                                                                                                                                                                        |
```
jorisvandenbossche, line 55: Small styling nit, but I would keep the `|` as they were (without all the extra whitespace). Markdown (at the github flavored one) doesn't require this to be rendered correctly, and give this is so wide, IMO it also doesn't provide any benefit in an editor, but rather makes it harder to read (since it now gets wrapped over multiple lines)

format-specs/geoparquet.md
jorisvandenbossche: What's the reason exactly to change this? 

(I am not necessarily opposed to be clear, just don't directly remember that we discussed this)

format-specs/geoparquet.md
```diff
@@ -50,57 +50,100 @@ Version of the geoparquet spec used, currently 0.3.0
...
+Users are recommended to store their data in longitude, latitude (OGC:CRS84 or not including the `crs` field) for it to work with the widest number of tools. Data that are more appropriately represented in particular projections may use an alternate coordinate reference system. We expect many tools will support alternate CRSes, but encourage users to check to ensure their chosen tool supports their chosen CRS.
+
+
+For implementations that operate entirely with longitude, latitude coordinates
+and are not CRS-aware or do not have easy access to CRS-aware libraries that can
+fully parse PROJJSON, it may be possible to infer that coordinates conform to
+the OGC:CRS84 CRS based on elements of the `crs` field.  For simplicity, Javascript
+object dot notation is used to refer to nested elements.
+
+The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` element is present:
```
jorisvandenbossche, line 131: Maybe not that important for this PR (can be done in a follow-up as well), but what do people think about moving the details of what the CRS repr of OGC:CRS84 and how to recognize this to a separate section towards the bottom of this file? So here we focus on the basic explanation of this metadata fields

format-specs/geoparquet.md
```diff
@@ -50,57 +50,100 @@ Version of the geoparquet spec used, currently 0.3.0
...
+* `"type"` = `"GeographicCRS"`
+* `coordinate_system.axis[0].unit` = `"degree"`
+* `coordinate_system.axis[1].unit` = `"degree"`
+* the values for `coordinate_system.axis[n].direction` are `"east"` and `"north"` (in either order, for n in [0,1])
+
+and at least one of the following are true:
+* `datum.id.authority` = `"EPSG"` and `datum.id.code` = `6326`
+* `datum_ensemble.id.authority` = `"EPSG"` and `datum_ensemble.id.code` = `6326`
+* `datum_ensemble.ellipsoid.semi_major_axis` = `6378137` and `datum_ensemble.ellipsoid.inverse_flattening` = `298.257223563`
+* `datum.ellipsoid.semi_major_axis` = `6378137` and `datum.ellipsoid.inverse_flattening` = `298.257223563`
```
jorisvandenbossche, line 146: The "checks" that are described here are for checking quite exact equivalence with EPSG:4326 or OGC:CRS84 (using the exact same datum definition), but based on https://github.com/opengeospatial/geoparquet/discussions/98 that might not necessarily be needed?

format-specs/geoparquet.md
brendan-ward: My intent was: "may" indicates reader may do so at their own peril (i.e., if you really don't care about ellipsoid); "should" indicates we recommend that the reader do so - which seems a bit strong when the ellipsoid of the long,lat coords is not defined.

rouault (approved):

format-specs/geoparquet.md
cholmes: I'm for removing the however and stating what it should be. Though perhaps we just say 'must use longitude, latitude with the WGS84 datum' to cover both OGC:CRS84 and EPSG:4326 (with axis order flipped).

cholmes (approved):
Looks good. I put in one small comment, but I'm also happy to merge this and then tackle that in a separate PR.

format-specs/geoparquet.md
```diff
@@ -50,57 +50,100 @@ Version of the geoparquet spec used, currently 0.3.0
...
+fully parse PROJJSON, it may be possible to infer that coordinates conform to
+the OGC:CRS84 CRS based on elements of the `crs` field.  For simplicity, Javascript
+object dot notation is used to refer to nested elements.
+
+The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` element is present:
+
+* `id.authority` = `"OGC"` and `id.code` = `"CRS84"`
+* `id.authority` = `"EPSG"` and `id.code` = `4326` (due to longitude, latitude ordering in this specification)
+
+The CRS is likely equivalent to OGR:CRS84 if all of the following are true:
```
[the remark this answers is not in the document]
cholmes: +1 - I think it's ok to even just remove this section. Like perhaps add it to some 'best practices' type document, but not have it in the main spec.

format-specs/geoparquet.md
jorisvandenbossche: > "should" indicates we recommend that the reader do so - which seems a bit strong when the ellipsoid of the long,lat coords is not defined.

It seems the last discussion seems to lean towards assuming WGS84? (which is still not a well defined ellipsoid without an epoch, but OK ;))

format-specs/geoparquet.md
jorisvandenbossche: +1 on removing the "however ..". So that if CRS is not available, readers can assume WGS84

examples/example_metadata.json
```diff
@@ -8,7 +8,111 @@
           180.0,
           83.6451
         ],
-        "crs": "GEOGCRS[\"WGS 84 (CRS84)\",ENSEMBLE[\"World Geodetic System 1984 ensemble\",MEMBER[\"World Geodetic System 1984 (Transit)\"],MEMBER[\"World Geodetic System 1984 (G730)\"],MEMBER[\"World Geodetic System 1984 (G873)\"],MEMBER[\"World Geodetic System 1984 (G1150)\"],MEMBER[\"World Geodetic System 1984 (G1674)\"],MEMBER[\"World Geodetic System 1984 (G1762)\"],MEMBER[\"World Geodetic System 1984 (G2139)\"],ELLIPSOID[\"WGS 84\",6378137,298.257223563],ENSEMBLEACCURACY[2.0]],CS[ellipsoidal,2],AXIS[\"geodetic longitude (Lon)\",east],AXIS[\"geodetic latitude (Lat)\",north],UNIT[\"degree\",0.0174532925199433],USAGE[SCOPE[\"Not known.\"],AREA[\"World.\"],BBOX[-90,-180,90,180]],ID[\"OGC\",\"CRS84\"]]",
+        "crs": {
+          "$schema": "https://proj.org/schemas/v0.4/projjson.schema.json",
```
brendan-ward, line 12: Note: for consistency, I used the full JSON as it is output in the example script, not the short version at the bottom of the spec.

cholmes:
Is this ready to merge? It's looking good to me. Thanks for all your awesome work on this @brendan-ward (and everyone else reviewing / suggesting).

brendan-ward:
The only thing outstanding is what part (if any) of the "OGC:CRS84 details" section to cut out or move to a different file (if so, where?).

cholmes:
Ah, right. I think the whole [OGC:CRS84 details](https://github.com/opengeospatial/geoparquet/blob/79eb2bcbb174f1ff94a75d921cf26fe849dbd48d/format-specs/geoparquet.md#ogccrs84-details) section could move out.

I'd probably just do something simple like just make an 'additional-info.md' file that sits right next the spec, in the same folder, and puts the crs details in it, maybe also the example file above. It could expand out in time, could also evolve to live somewhere else, like become more of a guide to geospatial for non-geo people, help explain various aspects of the spec.

I think it's also ok to just merge this and then we can do that one as a separate PR. I'd be happy to work on it.

jorisvandenbossche:
> Ah, right. I think the whole [OGC:CRS84 details](https://github.com/opengeospatial/geoparquet/blob/79eb2bcbb174f1ff94a75d921cf26fe849dbd48d/format-specs/geoparquet.md#ogccrs84-details) section could move out.

Given that it was moved to an additional section at the bottom, for me it is also fine to keep it there for now (we can move it another file later, if we prefer that), in light of getting this PR merged.

jorisvandenbossche (approved):

Issue #52, opened 2022-03-25
Should we recommend EPSG:4326 or something else?


cholmes:
This was discussed extensively in #25, but it feels worth revisiting. I think everyone feels good that the core recommendation is to use longitude, latitude in the WKB as the interoperability recommendation. The main question here is how we 'describe' that - use 4326 but then rely on the 'override' in our spec to put longitude first, or use something like OGC:84 that is less popular but actually describes things right.

Other points that were originally in #35:

- Do we want to continue recommending EPSG:4326? 
  - The [GDAL page](https://gdal.org/user/coordinate_epoch.html) on this topic says: *"The generic EPSG:4326 WGS 84 CRS is also considered dynamic, although it is not recommended to use it due to being based on a datum ensemble whose positional accuracy is 2 meters, but prefer one of its realizations, such as WGS 84 (G1762)"*
  - For example, QGIS now warns about this, see https://twitter.com/nyalldawson/status/1390118738251317254 for some context.

cayetanobv:
> This was discussed extensively in #25, but it feels worth revisiting. I think everyone feels good that the core recommendation is to use longitude, latitude in the WKB as the interoperability recommendation. The main question here is how we 'describe' that - use 4326 but then rely on the 'override' in our spec to put longitude first, or use something like OGC:84 that is less popular but actually describes things right.

As Geoparquet will follow WKB order of axis, and therefore the order of axis defined by any CRS will always be override, I agree with you to include something like you are proposing. I wrote this similar text to include at the end of crs section in spec:

```
Note that as indicated in [Coordinate axis order](link...) the order of the axis in Geoparquet is always x,y, so the EPSG:4326 (latitude, longitude) definition will be overridden. In this sense, the use of OGC:CRS84, which is a variant of EPSG:4326 but differs only in its coordinate order, could be a reasonable option despite being less popular.
```

> Other points that were originally in #35:
> 
> * Do we want to continue recommending EPSG:4326?
>   
>   * The [GDAL page](https://gdal.org/user/coordinate_epoch.html) on this topic says: _"The generic EPSG:4326 WGS 84 CRS is also considered dynamic, although it is not recommended to use it due to being based on a datum ensemble whose positional accuracy is 2 meters, but prefer one of its realizations, such as WGS 84 (G1762)"_
>   * For example, QGIS now warns about this, see https://twitter.com/nyalldawson/status/1390118738251317254 for some context.

[...]

PR #97, opened 2022-05-12
Require PROJJSON instead of WKT2:2019

### LINKED ISSUE #95: Use projjson instead of wkt2

brendan-ward:
Resolves #95 based on associated discussions in #90 and #52.

Supersedes #94 by incorporating a modification of that text here.

Supersedes #96 by requiring `crs` instead of allowing it to be optional, with associated text changes to express this.  PROJJSON for OGC:CRS84 and rules for inferring it are equivalent to #96.

This includes a minor formatting change of the table including crs but only the crs row changed content and position within the table.


format-specs/geoparquet.md
```diff
@@ -50,57 +50,116 @@ Version of the geoparquet spec used, currently 0.3.0
...
+
+The CRS must be provided in
+[PROJJSON](https://proj.org/specifications/projjson.html) format, which is a
+JSON encoding of
+[WKT2:2019 / ISO-19162:2019](https://docs.opengeospatial.org/is/18-010r7/18-010r7.html),
+which itself implements the model of
+[OGC Topic 2: Referencing by coordinates abstract specification / ISO-19111:2019](http://docs.opengeospatial.org/as/18-005r4/18-005r4.html).
+Apart from the difference of encodings, the semantics is intended to be exactly
+the same as WKT2:2019, and PROJJSON can be morphed losslessly from/into
+WKT2:2019.
```
tschaub, line 76: It looks like PROJJSON mentions [a few deviations](https://proj.org/specifications/projjson.html#deviations-with-the-wkt2-2019-specification) from WKT2:2019.  Maybe "exactly" and "losslessly" are too strong here.

How about "Apart from the difference of encodings, the semantics of PROJJSON are intended to match WKT2:2019, and a CRS in one encoding can generally be represented in the other"?

rouault: Regarding this, it could be wise, for greater interoperability, to discourage the use of BoundCRS objects, softly for top-level (as this maps to standard WKT2), and strongly for BoundCRS embedded in CompoundCRS (as this is a PROJ extension), for the PROJJSON CRS of GeoParquet.
BoundCRS objects are mostly the port of PROJ.4 +nadgrids/+towgs84/+geoidgrids constructs that are less necessary nowadays since PROJ can use its database to infer transformations between datums.
So the list of top-level objects suggested for use in PROJJSON would be:
- GeographicCRS 2D (or 3D if 3D data),
- ProjectedCRS 2D (note: ProjectedCRS 3D with ellipsoidal height is supported by PROJ, but still a bit of esoteric)
- CompoundCRS with horizontal part being a GeographicCRS 2D or a ProjectedCRS 2D and vertical part being a VerticalCRS
- and perhaps, GeodeticCRS with cartesian geocentric coordinate system (X,Y,Z) (probably of marginal use)

format-specs/geoparquet.md
```diff
@@ -50,57 +50,116 @@ Version of the geoparquet spec used, currently 0.3.0
...
-        AXIS["geodetic latitude (Lat)",north,
-            ORDER[2],
-            ANGLEUNIT["degree",0.0174532925199433]],
-    USAGE[
-        SCOPE["Not known."],
-        AREA["World."],
-        BBOX[-90,-180,90,180]],
-    ID["OGC","CRS84"]]
-```
+The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id`
```
cholmes, line 145: Thanks for putting this together. I have some thoughts on it, but seemed to be a bigger scope than just a line in a PR, getting into if we might be able to tweak projjson. See https://github.com/opengeospatial/geoparquet/discussions/98

rouault:
Still seing this open/ I'm confused: I thought that PROJJSON was already part of GeoParquet v0.4 !  And existing GDAL versions (latest 3.5.x and 3.6.0) that write GeoParquet v0.4 do write PROJJSON !
I believe this should be prioriterized to the next release. Or I'll have to downgrade the OGR GeoParquet driver to write WKT2.

brendan-ward:
@rouault sorry, my mistake for not closing this.  It was superseded by #96, which did indeed bring in PROJJSON support.  The distinction here was about whether or not `crs` is required and set explicitly to null when CRS is not known.  I think this came out of an earlier meeting where folks wanted to see both options.

PR #104, opened 2022-05-25
Bump json schema version to 0.4.0

kylebarron:
### Change list

- Bumps version in JSON Schema to `0.4.0`
- Changes from regex to const string matching. This is the same approach [used in the STAC spec](https://github.com/radiantearth/stac-spec/blob/789e90fba4c7119bc925d26f75f47bacf44c91e6/item-spec/json-schema/item.json#L94-L96). The thought being that this schema definition is tied to a single version, and that if we publish a patch version, there will be a new json schema for 0.4.1.
- Updates example file metadata to 0.4.0

We might also want to point people to a tagged URL for the json schema definition, i.e.
```
https://raw.githubusercontent.com/opengeospatial/geoparquet/v0.4.0/format-specs/schema.json
```


kylebarron:
Good to see that CI works 😄

cholmes (approved):

jorisvandenbossche (approved):

PR #108, opened 2022-05-26
add R examples and geoarrow

yeelauren:
Added `geoarrow` package for R in readme and a minimal example with plotting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,5 +82,6 @@ available on Google Cloud Storage.
 Currently known libraries that can read and write GeoParquet files:
 
 * [GeoPandas](https://geopandas.org/en/stable/docs/user_guide/io.html#apache-parquet-and-feather-file-formats) (Python)
+* [geoarrow](https://github.com/paleolimbot/geoarrow) (R)
 * [sfarrow](https://wcjochem.github.io/sfarrow/index.html) (R)
 * [GDAL/OGR](https://gdal.org/drivers/vector/parquet.html) (C++, bindings in several languages)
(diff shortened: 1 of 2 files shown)


cholmes:
Awesome! Excited to add this. It looks like you had a markdown formatter going so this PR has a lot more changes than just the couple of new lines. Could you make it so it's just a PR on the new lines?

I don't think we're opposed to markdown formatting improvements, but it'd be best to put them in their own, clean PR so the changes are clear. I do think we should move towards some automated formatting or at least lint checking for markdown, so it is all consistent.

kylebarron:
probably good to cc @paleolimbot directly too for 👍

cholmes (approved):

cholmes:
Sorry for the delay on this, but looks great - thanks for your contribution @yeelauren!

cholmes:
Ugh, I merged this but it looks like it lead to a lint failure - I apologize. I have to run right now, but I'll try to investigate soon. It'd be good to run our linter on branches and not merge if it's failing, so we hit the errors before merging instead of after.

kylebarron:
Lint is already set up to run on every pull request:
https://github.com/opengeospatial/geoparquet/blob/d00ad6291ff8f751f8210dec7b6c747c5920a5f9/.github/workflows/lint.yml#L7

On the first commit on this PR lint failed, but then not sure why no later commits show that CI was run at all

jorisvandenbossche:
This also makes me wonder: do we somehow want to keep track for each implementation we list which version they support? 
Because I see that `geoarrow` is still at 0.3.0, and doesn't yet write 0.4.0 files.

Issue #79, opened 2022-04-22
How should metadata be written in a partitioned dataset?


kylebarron:
So far the spec has only covered single-file Parquet data. However Parquet also supports saving as a "dataset", where there are several Parquet files in a folder structure. In this case, how should geospatial metadata be stored? There's a Parquet [best practice](https://arrow.apache.org/docs/python/parquet.html#writing-metadata-and-common-medata-files) that writes `_common_metadata` and `_metadata` sidecar files to the root of the folder structure, but that's not part of the actual Parquet specification.

If I understand correctly, the `geo` metadata would automatically be included in the `_common_metadata` file, and additionally statistics are stored in the `_metadata` file, which is relevant for #13 

Should this be part of the geoparquet spec? Should it be a "best practice" that we document?

kylebarron:
Adding here for discussion from https://github.com/opengeospatial/geoparquet/discussions/101#discussioncomment-2784284; we should clarify what the bounding box represents.

Is it possible for each file's metadata to contain only _its_ bounding box, but for the `_metadata` file to contain metadata for the _entire dataset's_ bounding box? Otherwise, we should probably have the same bounding box in every file, representing the entire dataset?

cholmes:
Where do we stand on this in relation [latest discussion](https://github.com/opengeospatial/geoparquet/discussions/122) pushing for stac 1.0.0-beta.1 sooner rather than later. Should we attempt to put something in there? Has anyone experimented with this and have a good recommendation here?

cholmes:
Do we want to do something here for 1.1? We've now got 'spatial optimizations' ready to go for 1.1, so it seems like a good time to get this to completion to. It sounds like libraries do make use of the `_metadata` sidecar files, in the comment referenced above Even said 

> [GDAL] uses the arrow-dataset library helpers. When it finds _metadata, it uses the arrow::dataset::ParquetDatasetFactory class, which avoids reading each fragment file for most operations (except for GetExtent() since there's no way of having the global extent without iterating over each fragment metadtaa). When it doesn't find _metadata, it uses the arrow::dataset::FileSystemDatasetFactory class (not sure if that one needs to open each file)'

Do we want to explicitly say that the _metadata sidecar sets the extent of the whole dataset? Or just leave things as they are?

kylebarron:
> _metadata sidecar sets the extent of the whole dataset

[...]

PR #113, opened 2022-06-14
BigQuery scripts to import/export GeoParquet files

alasarr:
This PR allows:
 - Export to GeoParquet a dataset (via SQL query) from BigQuery. 
 - Import in BigQuery a dataset.

It doesn't aim to be a production-ready solution, in the future, it will be supported natively by BigQuery. However, in the meantime, it provides a way to work with GeoParquet and BigQuery.

Convert a SQL query to parquet:

```bash
poetry run python bigquery_to_parquet.py \
    --input-query "SELECT * FROM carto-do-public-data.carto.geography_usa_blockgroup_2019" \
    --primary-column geom \
    --output geography_usa_blockgroup_2019 
```

Upload a parquet file or folder to BigQuery:
```bash
poetry run python parquet_to_bigquery.py \
    --input geography_usa_blockgroup_2019 \
    --output "cartodb-gcp-backend-data-team.alasarr.geography_usa_blockgroup_2019"
```

I've extracted to a global module some of the code generated at https://github.com/opengeospatial/geoparquet/pull/87

### DIFF
--- a/scripts/bigquery_to_parquet.py
+++ b/scripts/bigquery_to_parquet.py
@@ -1,0 +1,90 @@
+
+
+import click
+import sys
+import pyarrow.parquet as pq
+import geopandas as gpd
+
+from encoder import AVAILABLE_COMPRESSIONS, Edges, PathType, geopandas_to_arrow
+from pathlib import Path
+from google.cloud import bigquery
+
+MODES = ["FILE", "FOLDER"]
+
+def read_gdf(input_query: str, primary_column: str):
+    client = bigquery.Client()
+    df = client.query(input_query).to_dataframe()
+    df[primary_column] = gpd.GeoSeries.from_wkt(df[primary_column])
+    return gpd.GeoDataFrame(df, geometry=primary_column, crs="EPSG:4326")
+
+@click.command()
+@click.option(
+    "-q",
+    "--input-query",
+    type=str,
+    help="SQL query of the data to export",
+    required=True,
+)
+@click.option(
+    "--primary-column",
+    type=str,
+    help="The primary column name with geometry data",
+    required=True,
+)
+@click.option(
+    "-o",
+    "--output",
+    type=PathType(file_okay=True, dir_okay=True, writable=True),
+    help="Path to output",
+    required=True,
+)
+@click.option(
+    "-m",
+    "--mode",
+    type=click.Choice(MODES, case_sensitive=False),
+    help="Mode to use FILE or FOLDER",
+    default="FOLDER",
+    show_default=True
+)
+@click.option(
+    "--compression",
+    type=click.Choice(AVAILABLE_COMPRESSIONS, case_sensitive=False),
+    default="SNAPPY",
+    help="Compression codec to use when writing to Parquet.",
+    show_default=True,
+)
+@click.option(
+    "--partition-size",
+    type=int,
+    default=5000,
+    help="Number of records per partition. Ignored if --single-file is provided.",
+    show_default=True,
+)
+def main(input_query: str, primary_column: str, output: Path, mode: str, compression: str , partition_size: int):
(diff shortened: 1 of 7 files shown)


scripts/pyproject.toml
```diff
@@ -6,12 +6,14 @@ authors = []
 license = "MIT"
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = ">=3.8,<3.11"
```
alasarr, line 9: Google BigQuery requires this

scripts/bigquery_to_parquet.py
```diff
@@ -0,0 +1,90 @@
...
+from pathlib import Path
+from google.cloud import bigquery
+
+MODES = ["FILE", "FOLDER"]
+
+def read_gdf(input_query: str, primary_column: str):
+    client = bigquery.Client()
+    df = client.query(input_query).to_dataframe()
+    df[primary_column] = gpd.GeoSeries.from_wkt(df[primary_column])
+    return gpd.GeoDataFrame(df, geometry=primary_column, crs="EPSG:4326")
```
kylebarron, line 18: This is a big, unnecessary bottleneck. Assuming that you're using recent versions of the python bigquery client library, bigquery downloads the data as Arrow. So in effect this makes two extra copies of the entire dataset: first to copy into Numpy/GEOS memory and again to copy back into Arrow memory. If the only purpose is to write to geoparquet, you can just parse the WKT to WKB, keep it in Arrow memory, and manually add metadata

kylebarron: (This would require a little more custom code than the existing approach that reuses the existing `geopandas_to_arrow` function, but would be worth it if performance is important)

scripts/bigquery_to_parquet.py
```diff
@@ -0,0 +1,90 @@
...
+    print("Finished reading", file=sys.stderr)
+    print("Starting conversion to Arrow", file=sys.stderr)
+    arrow_table = geopandas_to_arrow(gdf, Edges.SPHERICAL)
+    print("Finished conversion to Arrow", file=sys.stderr)
+
+    print("Starting write to Parquet", file=sys.stderr)
+
+    if mode.upper() == 'FOLDER':
+        # We need to export to multiple files, because a single file might hit bigquery limits (UDF out of memory). https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-parquet
+        pq.write_to_dataset(arrow_table, root_path=output, partition_cols=['__partition__'], compression=compression)
```
kylebarron, line 83: We should still have a discussion on best practices for partitioned datasets. https://github.com/opengeospatial/geoparquet/issues/79 Specifically around whether the `_metadata` file is required/suggested/etc.

scripts/pyproject.toml
```diff
@@ -6,12 +6,14 @@ authors = []
...
 pyarrow = "^7.0.0"
 geopandas = "^0.10.2"
 pygeos = "^0.12.0"
 pandas = "^1.4.2"
 click = "^8.1.2"
+google-cloud-bigquery = "^3.2.0"
+db-dtypes = "^1.0.2"
```
kylebarron, line 16: Is this used? It doesn't appear to be imported

scripts/bigquery_to_parquet.py
alasarr: Good point, will follow the conversation there

scripts/pyproject.toml
alasarr: I added it because google-cloud-bigquery complained about it by asking to install it, so I did

cholmes:
Closing this, as we'll move it over to https://github.com/geoparquet/bigquery-converter and keep the main geoparquet repo more focused on the spec.

PR #121, opened 2022-10-24
Include suggestion about feature identifiers

### LINKED ISSUE #120: Feature identifiers
Has there been discussion around including an `id_column` or something similar in the file metadata?  I think it would assist in round-tripping features from other formats if it were known which column represented the feature identifier.

It looks like GDAL has a [FID layer creation option](https://gdal.org/drivers/vector/parquet.html#layer-creation-options).  But I'm assuming that the information about which column was used when writing would be lost when reading from the parquet file (@rouault would need to confirm).

I grant that this doesn't feel "geo" specific, and there may be existing conventions in Parquet that would be appropriate.

tschaub:
Based on discussion in #120, this adds documentation that suggests adding custom file key/value metadata for things like feature identifiers.

I didn't use language like "primary key" as suggested in previous discussion.  I think these feature identifiers typically represent "external" identifiers instead of something like a primary key in a database.

If others think that the extra language risks misinterpretation or is not useful, I'm not attached to having it in.

Fixes #120.

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -166,6 +166,12 @@ This follows the GeoJSON specification ([RFC 7946, section 5](https://tools.ietf
 
 ### Additional information
 
+#### Feature identifiers
+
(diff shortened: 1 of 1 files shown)


cholmes (approved):

jorisvandenbossche (approved):
Looks good to me to include this note

Issue #129, opened 2022-10-25
Potential JSON Schema issues


m-mohr:
Some potential improvements and/or issues in the JSON Schema:

- [x] `$defs` keyword in the schema was only added in draft 2019-09 (see also #127 )
- [x] `additionalProperties: true` is the default and could be removed
- [x] `"patternProperties":`  with a pattern `".*"` is basically equivalent to additionalProperties and thus can be simplified a bit.
- [x] `"enum": ["WKB"]` is equivalent to `"const": "WKB"`, the same applies to `"enum": ["counterclockwise"]`
- [x] PROJJSON has a new schema version, 0.5
- [x] The bbox schema looks weird. Are more than 4 values allowed? Right now it would allow [0,1,2,3,"asdsad"]. Could probably also be simplified using min/maxItems... I think you were looking for something like `"items": {"type": "number"}, "minItems": 4`
- [x] By the way, the spec says `double` for epoch, but this is not enforced in the schema and could be an int (or should this be "number" in the spec?)
- [x] I understand the spec that you can either specify an array of geometry types **or** a single geometry type as string **or** `Unknown` (so always as a string, not in an array). If this is the case, the schema doesn't check it. `["Polygon", "Unknown"]` is valid.
- [x] The Z suffix to the geometry types is not part of the JSON Schema it seems

Happy to work on PRs, just let me know whether all this is correctly captured.

jorisvandenbossche:
I am not very familiar with JSON schema details, but that all sounds correct from the geoparquet side.

> The bbox schema looks weird. Are more than 4 values allowed? Right now it would allow [0,1,2,3,"asdsad"]. Could probably also be simplified using min/maxItems... I think you were looking for something like `"items": {"type": "number"}, "minItems": 4`

I am currently updating this field in https://github.com/opengeospatial/geoparquet/pull/88, so your feedback there is certainly welcome! (we are basically already doing something with items and minItems now, just spelling out the 4 or 6 values explicitly)


> `"enum": ["WKB"]` is equivalent to `"const": "WKB"`, the same applies to `"enum": ["counterclockwise"]`

Although it is not the case right now, but we are planning to add more options in the future, so can maybe keep it like that? (although also the enum wouldn't allow any other value right now, so it wouldn't really matter)

m-mohr:
Okay, I'll draft a PR.

> I am currently updating this field in #88, so your feedback there is certainly welcome!

Done.

> 
[...]

PR #131, opened 2022-10-25
Update JSON Schema and data types #128, 129

### LINKED ISSUE #128: crs data type: JSON object
The CRS data type says "JSON object", which could be understood as object serialized as JSON (i.e. a string), so this should probably just say "object"?

m-mohr:
Fixes #128 and #129.

I've added comments below with some further details.

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -31,7 +31,7 @@ All file-level metadata should be included under the "geo" key in the parquet me
 | ------------------ | ------ | -------------------------------------------------------------------- |
 | version     		 | string | **REQUIRED** The version of the geoparquet metadata standard used when writing.   |
 | primary_column     | string | **REQUIRED** The name of the "primary" geometry column.                |
-| columns            | Map<key, [Column Metadata](#column-metadata)> | **REQUIRED** Metadata about geometry columns, with each key is the name of a geometry column in the table. |
+| columns            | Map\<str, [Column Metadata](#column-metadata)> | **REQUIRED** Metadata about geometry columns. Each key is the name of a geometry column in the table. |
 
 At this level, additional implementation-specific fields (e.g. library name) are allowed, and thus readers should be robust in ignoring those.
 
@@ -52,13 +52,13 @@ Each geometry column in the dataset must be included in the columns field above 
 
 | Field Name | Type | Description |
 | --- | --- | --- |
-| encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported. |
-| geometry_type | string or \[string] | **REQUIRED** The geometry type(s) of all geometries, or 'Unknown' if they are not known. |
-| crs | JSON object | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) JSON object representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude based on the WGS84 datum, and CRS-aware implementations should assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). |
-| orientation | string | **OPTIONAL** Winding order of exterior ring of polygons. If present must be 'counterclockwise'; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
-| edges | string | **OPTIONAL** Name of the coordinate system for the edges. Must be one of 'planar' or 'spherical'. The default value is 'planar'. |
+| encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only `WKB` is supported. |
+| geometry_type | string or \[string] | **REQUIRED** The geometry type(s) of all geometries, or `Unknown` if they are not known. |
+| crs | object | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) JSON object representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude based on the WGS84 datum, and CRS-aware implementations should assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). |
+| orientation | string | **OPTIONAL** Winding order of exterior ring of polygons. If present must be `counterclockwise`; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
+| edges | string | **OPTIONAL** Name of the coordinate system for the edges. Must be one of `planar` or `spherical`. The default value is 'planar'. |
 | bbox | \[number] | **OPTIONAL** Bounding Box of the geometries in the file, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
-| epoch | double | **OPTIONAL** Coordinate epoch in case of a dynamic CRS, expressed as a decimal year. |
+| epoch | number | **OPTIONAL** Coordinate epoch in case of a dynamic CRS, expressed as a decimal year. |
 
 #### crs
 
--- a/format-specs/schema.json
+++ b/format-specs/schema.json
@@ -3,6 +3,7 @@
   "title": "GeoParquet",
   "type": "object",
   "description": "Parquet metadata included in the geo field.",
+  "required": ["version", "primary_column", "columns"],
   "properties": {
     "version": {
       "type": "string",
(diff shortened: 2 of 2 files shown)


format-specs/geoparquet.md
```diff
@@ -52,13 +52,13 @@ Each geometry column in the dataset must be included in the columns field above
...
+| crs | object | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) JSON object representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude based on the WGS84 datum, and CRS-aware implementations should assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). |
+| orientation | string | **OPTIONAL** Winding order of exterior ring of polygons. If present must be `counterclockwise`; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
+| edges | string | **OPTIONAL** Name of the coordinate system for the edges. Must be one of `planar` or `spherical`. The default value is 'planar'. |
 | bbox | \[number] | **OPTIONAL** Bounding Box of the geometries in the file, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
-| epoch | double | **OPTIONAL** Coordinate epoch in case of a dynamic CRS, expressed as a decimal year. |
+| epoch | number | **OPTIONAL** Coordinate epoch in case of a dynamic CRS, expressed as a decimal year. |
```
m-mohr, line 61: This is allowed to be an integer, right?

format-specs/schema.json
```diff
@@ -11,95 +12,84 @@
...
     }
   },
-  "additionalProperties": true,
-  "required": ["version", "primary_column", "columns"],
-  "$defs": {
+  "definitions": {
     "geometry_type": {
       "type": "string",
-      "enum": ["Point", "LineString", "Polygon", "MultiPoint", "MultiLineString", "MultiPolygon", "GeometryCollection", "Unknown"]
+      "pattern": "^(GeometryCollection|(Multi)?(Point|LineString|Polygon))( Z)?$"
```
m-mohr, line 92: Added Z and removed Unknown (see above). Could also be a long enum, but thought it might be simpler to use a regexp. We can switch back though if that's preferred. I assume `GeometryCollection Z` is allowed?

format-specs/schema.json
```diff
@@ -11,95 +12,84 @@
...
-                  "description": "The maximum constant latitude line that bounds the rectangle (ymax)."
-                }
-              ]
+                "uniqueItems": true
+              }
+            ],
+            "description": "The geometry type(s) of all geometries, or 'Unknown' if they are not known."
+          },
+          "crs": {
+            "$ref": "https://proj.org/schemas/v0.5/projjson.schema.json"
```
m-mohr, line 50: Removed the `null` schema for CRS, becuase null is not mentioned in the spec.

format-specs/schema.json
```diff
@@ -11,95 +12,84 @@
...
+            "type": "string",
+            "enum": ["planar", "spherical"],
+            "description": "Name of the coordinate system for the edges. Must be one of 'planar' or 'spherical'. The default value is 'planar'."
+          },
+          "orientation": {
+            "type": "string",
+            "const": "counterclockwise",
+            "description": "Winding order of exterior ring of polygons; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order."
+          },
+          "bbox": {
```
m-mohr, line 62: See also #88

format-specs/schema.json
kylebarron: I think we may have had a discussion about this a while ago, and it seems like we had a preference for having a literal `null` (like we require a literal `Unknown` for geometry type) instead of having the key be missing.

kylebarron (commented):
A lot of good changes (the STAC experience shows!) though I think it would be easier to review/discuss if split into a few different PRs.

format-specs/schema.json
kylebarron: Do we need to declare `Z` in the geometry type? It might make it a little harder on the writer side

format-specs/geoparquet.md
jorisvandenbossche: It should be a float, in theory, but I suppose implementations won't care if it's rounded?

format-specs/schema.json
m-mohr: That's what the spec states at least... so more a discussion for the general issue tracker?

format-specs/schema.json
m-mohr: Okay, I can change it back and update the spec if that's confirmed.

format-specs/geoparquet.md
m-mohr: We could in theory enforce a float, but I've never seen this before in JSON Schema...

format-specs/schema.json
jorisvandenbossche: Yes, having

jorisvandenbossche: Yes, having `"crs": null` is different from a missing "crs" key

It is mentioned here in the spec:

https://github.com/opengeospatial/geoparquet/blob/811d52bdb57814dbf0c13049481b8f3b2367b31d/format-specs/geoparquet.md?plain=1#L85

m-mohr:
@kylebaron The issue is the indentation change for patternProperties/additionalProperties. Makes it a bit hard to compare in the underwhelming GitHub diff tool. Maybe open it in a another diff tool?

format-specs/schema.json
jorisvandenbossche: Yes, let's indeed discuss that in a separate issue in case we want to reconsider this

m-mohr: See #132

format-specs/geoparquet.md
m-mohr: Would be something like this:
```json
{
  "type": "number",
  "not": {
  	"type": "integer"
  }
}
```

kylebarron:
> @kylebaron The issue is the indentation change for patternProperties/additionalProperties.

I was referring more towards:
- Whether to declare Z in the geom type
- Whether crs should allow `null`
- Exactly how `bbox` should be defined in the schema

Could each be their own PR with their own discussion

format-specs/geoparquet.md
tschaub: I would avoid gymnastics like this.  Instead stick with JSON types when using JSON.

rouault: number, allowing integer or floating-point, should be just fine

format-specs/schema.json
tschaub: This change also removes "Unknown" (which is mentioned [in the text](https://github.com/opengeospatial/geoparquet/blob/811d52bdb57814dbf0c13049481b8f3b2367b31d/format-specs/geoparquet.md#column-metadata) of the spec).

I'll open a separate issue for this, but I wonder if folks would be in support of requiring an array of strings instead of a string or array of strings.  Makes life a bit simpler where this is parsed into a struct and a specific type is useful.

m-mohr: As mentioned in my first comment: I moved Unknown (i.e. added as a separate subschema) in the corresponding property.
Reason: It reads as if Unknown is not meant to be used together with other types.

I was also wondering about the array vs string thing and would also say that just always using an array makes things easier (except that it requires a bit more logic to avoid things like ["Point", "Unknown"])

format-specs/geoparquet.md
m-mohr: Yeah, doesn't even work because 7.0 is a valid integer according to JSON schema. :-)

format-specs/schema.json
```diff
@@ -11,95 +12,92 @@
...
+            "oneOf":[
+              {
+                "description": "2D bbox consisting of (xmin, ymin, xmax, ymax)",
+                "minItems":4,
+                "maxItems":4
+              },
+              {
+                "description": "3D bbox consisting of (xmin, ymin, zmin, xmax, ymax, zmax)",
+                "minItems":6,
+                "maxItems":6
```
tschaub, line 85: I wouldn't go this far.  The [GeoJSON spec](https://www.rfc-editor.org/rfc/rfc7946#section-3.1.1) doesn't require (only recommends) 2 or 3 dimensions.

Instead, I would go with this:
```json
"bbox": {
  "type": "array",
  "minItems": 4,
  "items": {
    "type": "number"
  }
}
```

m-mohr: Also fine with me, I just reduced what has been discussed in #88...

jorisvandenbossche: Let's discuss bbox in #88 (maybe leave it as is here so it's not in the diff?)

format-specs/schema.json
tschaub: > As mentioned in my first comment: I moved Unknown (i.e. added as a separate subschema) in the corresponding property. Reason: It reads as if Unknown is not meant to be used together with other types.

Ah, I should have read more carefully.

> I was also wondering about the array vs string thing and would also say that just always using an array makes things easier (except that it requires a bit more logic to avoid things like ["Point", "Unknown"])

Right, I would use `[]` to signify an unknown geometry type.  See #133.

m-mohr:
I'll rebase this after #88 and #140 have been merged.

m-mohr:
Updated to reflect the latest changes.

m-mohr:
Two additional changes, see also https://github.com/opengeospatial/geoparquet/pull/141#issuecomment-1338197861:

1. Replaced `patternProperties` pattern ".*" with ".+" to not allow empty column names (see #153)
2. Don't allow additional columns that are invalid

format-specs/geoparquet.md
```diff
@@ -58,15 +58,15 @@ Version of the GeoParquet spec used, currently 0.4.0
...
-| bbox | \[number] | **OPTIONAL** Bounding Box of the geometries in the file, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
-| epoch | number | **OPTIONAL** Coordinate epoch in case of a dynamic CRS, expressed as a decimal year. |
+| Field Name     | Type         | Description |
+| -------------- | ------------ | ----------- |
+| encoding       | string       | **REQUIRED** Name of the geometry encoding format. Currently only `"WKB"` is supported. |
+| geometry_types | \[string]    | **REQUIRED** The geometry types of all geometries, or an empty array if they are not known. |
+| crs            | object\|null | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) object representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude based on the WGS84 datum, and CRS-aware implementations should assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). |
```
tschaub, line 65: It is stated below, but since `null` is being added here, I think the text should also describe what `null` means.  Something like "a value of `null` indicates that the CRS is undefined or unknown."

m-mohr:
```suggestion
| crs            | object\|null | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) object representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude based on the WGS84 datum, and CRS-aware implementations should assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). A value of `null` indicates that the CRS is undefined or unknown. |
```

tschaub (approved):

m-mohr:
Wasn't this meant to wait for #141 ?

tschaub:
My bad.  I thought I understood that this was going to come first.  I had just read the comment here https://github.com/opengeospatial/geoparquet/pull/141#issuecomment-1338197861 and thought that the tests were going to be adapted after this was in.

@jorisvandenbossche - Apologies for complicating things if the tests don't pass with these changes.

jorisvandenbossche:
No problem! The tests are also still passing, so no complication at all ;)

Issue #123, opened 2022-10-24
Example data to test implementations


cholmes:
One idea that @jorisvandenbossche suggested is that we should have a set of data that shows the range of the specification, that implementors can use to make sure they're handling right, and which could be the basis of 'integration' testing.

This would include geoparquet files that have a variety of projections, geometry types (including multiple geometry types as in #119), plus things like multiple geometry columns, different `edges` values, etc. It could also be good to have a set of 'bad' files that can also be tested against.

cholmes:
There was also work on this in #204 for the new geometry encodings. We should figure out if we want to try to do more for 1.1

PR #134, opened 2022-11-07
Add test data covering various options in the spec

jorisvandenbossche:
Related to https://github.com/opengeospatial/geoparquet/issues/123

This is incomplete (more parameters should be added) and still draft (the script should be cleaned-up, ensure to add this to the CI to check the generated files, validate with json schema, etc). But wanted to open a PR already to see where and how we want to go with this.


cholmes:
@jorisvandenbossche - do you want to try to get this updated for the 1.1 release? It sounds like it's complementary to #134, right?

jorisvandenbossche:
Yes, have to update it follow the same changes I made in https://github.com/opengeospatial/geoparquet/pull/204

PR #140, opened 2022-11-07
Array of geometry types

### LINKED ISSUE #132: Declare `Z` in the geometry type?
Do we need to declare `Z` in the geometry type? It might make it a little harder on the writer side

_Originally posted by @kylebarron in https://github.com/opengeospatial/geoparquet/pull/131#discussion_r1004712064_

### LINKED ISSUE #133: Geometry types
Currently, the `geometry_type` metadata property can be a string or an array of strings.  I'm wondering if this flexibility is really necessary.  When parsing this metadata into a struct, it can make things simpler to have a single type for the value.  

In addition, instead of the string `"Unknown"`, an empty array could represent unknown geometry types.

If an array sounds good, I think the name `geometry_types` also makes sense.  Basically, I'm wondering if there would be support for this instead:
```json
{
  "geometry_types": {
    "description": "The geometry types of all geometries, or an empty array if they are not known.",
    "type": "array",
    "items": {
      "type": "string",
      "pattern": "^(GeometryCollection|(Multi)?(Point|LineString|Polygon))( Z)?$"
    }
  }
}
```

tschaub:
This renames the `geometry_type` property to `geometry_types` and updates the JSON type to be an array of strings (instead of a string or array of strings).  The spec language has been updated to reflect that an empty array indicates that the geometry types are not known.  The schema now enforces that the items in the list are one of the expected types and allows the ` Z` suffix ~for everything except `GeometryCollection`~.

I updated the `example.parquet` file to use `geometry_types` instead of `geometry_type`.  (I followed the readme, but am struggling to run the validator, so admit I haven't yet done that.)

I bumped the schema version to `0.5.0-dev`.  Ideally this would happen as part of the process for releasing a tag (create release branch, update version, create tag, generate release, bump version back to `X.Y.Z-dev`, merge branch).

Fixes #132.
Fixes #133.


rouault:
> allows the ` Z` suffix for everything except `GeometryCollection`.

why not GeometryCollection ?  "GeometryCollection Z" is a valid WKB type

tschaub:
> why not GeometryCollection ? "GeometryCollection Z" is a valid WKB type

My bad.  I removed the language about this.

rouault (approved):

tschaub:
Should the `validate-examples` CI job be updated to use the validator and schema from the repo instead of from pypi?

kylebarron:
> Should the `validate-examples` CI job be updated to use the validator and schema from the repo instead of from pypi?

If I'm reading it correctly I think it is being installed from this repo

https://github.com/opengeospatial/geoparquet/blob/134ebef5f77360d906512503088295b00a9af329/.github/workflows/scripts.yml#L23

The `.` is referring to this repo, and `--no-binary geoparquet_validator` is I think just telling `pip` to force using the source instead of a binary wheel. Not 100% sure why we'd need `--no-binary` in any case

kylebarron:
Oh @tschaub you have to change this line https://github.com/tschaub/geoparquet/blob/9f94199ea2550ad4cf97b4923133afc55f4bc24c/format-specs/schema.json#L85

format-specs/geoparquet.md
```diff
@@ -110,26 +110,24 @@ The axis order of the coordinates in WKB stored in a geoparquet follows the de f
...
 "GeometryCollection".
 
 In addition, the following rules are used:
 
 - In case of 3D geometries, a " Z" suffix gets added (e.g. "Point Z").
-- The value can be a single string or an array of strings in case multiple
-  geometry types are present (e.g. ["Polygon", "MultiPolygon"]).
-- Additionally the value "Unknown" is accepted to explicitly signal that the
-  geometry type is not known.
+- A list of values indicates that multiple geometry types are present (e.g. ["Polygon", "MultiPolygon"]).
```
jorisvandenbossche, line 123:
```suggestion
- A list of multiple values indicates that multiple geometry types are present (e.g. ["Polygon", "MultiPolygon"]).
```

? (since we are now always using a list, this otherwise is a bit misleading?)

format-specs/geoparquet.md
```diff
@@ -110,26 +110,24 @@ The axis order of the coordinates in WKB stored in a geoparquet follows the de f
...
+- An empty array explicitly signals that the geometry types are not known.
 
 It is expected that this field is strictly correct. For
 example, if having both polygons and multipolygons, it is not sufficient to
 specify "MultiPolygon", but it is expected to specify
```
jorisvandenbossche, line 128:
```suggestion
specify ["MultiPolygon"], but it is expected to specify
```

format-specs/geoparquet.md
tschaub: Sounds good.  Updated in the latest commit.

jorisvandenbossche (approved):

format-specs/geoparquet.md
tschaub: Done.

tschaub:
> > Should the `validate-examples` CI job be updated to use the validator and schema from the repo instead of from pypi?
> 
> If I'm reading it correctly I think it is being installed from this repo
> 
> https://github.com/opengeospatial/geoparquet/blob/134ebef5f77360d906512503088295b00a9af329/.github/workflows/scripts.yml#L23
> 
> The `.` is referring to this repo, and `--no-binary geoparquet_validator` is I think just telling `pip` to force using the source instead of a binary wheel. Not 100% sure why we'd need `--no-binary` in any case. `geoparquet-validator` [isn't even pushed to pypi](https://pypi.org/project/geoparquet_validator/), so it can't be installing from there.

Thanks for the clarification.  I missed the `.`.

format-specs/geoparquet.md
```diff
@@ -53,7 +53,7 @@ Each geometry column in the dataset must be included in the columns field above
 | Field Name | Type | Description |
 | --- | --- | --- |
 | encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only 'WKB' is supported. |
-| geometry_type | string or \[string] | **REQUIRED** The geometry type(s) of all geometries, or 'Unknown' if they are not known. |
+| geometry_types | \[string] | **REQUIRED** The geometry types of all geometries, or an empty array if they are not known. |
```
m-mohr, line 56: If the field can be empty, why require it anyway?

format-specs/schema.json
```diff
@@ -25,20 +25,13 @@
...
-                  "items": {
-                    "$ref": "#/$defs/geometry_type"
-                  },
-                  "uniqueItems": true
-                }
-              ],
-              "description": "The geometry type(s) of all geometries, or 'Unknown' if they are not known."
+            "geometry_types": {
+              "description": "The geometry types of all geometries, or an empty array if they are not known.",
+              "type": "array",
```
m-mohr, line 30: I'd propose to add "uniqueItems" set to `true` so that it can't be `["Point", "Point"]` :-)

format-specs/geoparquet.md
m-mohr: Also, the description should probably mention that it's a set of values (i.e. unique items in the array - see also the other comment)

format-specs/schema.json
tschaub: Good point.  Added in the latest commit.

format-specs/geoparquet.md
tschaub: If it is optional, I assume we wouldn't assign any different meaning to `{}` and `{"geometry_types": []}`.  So then do we want `"minItems": 1`?

I like the idea of avoiding two ways to say the same thing.  No strong preference between optional with min length of one vs required.  Though it seems slightly more explicit to say that `[]` means "unknown" (whereas the absence of the property might either mean "unknown" or "I forgot" - not sure if there is any significant difference between the two though).

tschaub: I agree with the point I think you are making though.  If it isn't required that the geometry types be known, why require the geometry types in the metadata?

m-mohr: Yeah, if it is optional, minItems should be set to 1, indeed. I have a slight tendency for not requiring it as for me an empty array reads more as "no type", but requiring it makes people think (if implementations don't set a default such as  `write_geoparquet(geometry_types = [], ...)`

jorisvandenbossche: Some discussion on this topic happened in the original issue https://github.com/opengeospatial/geoparquet/issues/41. The idea was indeed that we wanted to avoid two ways to say the same thing (not specified if optional, vs `geometry_type="Unknown"`. 

And I think the rationale for making it required was indeed to have people think about it / recommend to specify a useful value (instead of just leaving out): https://github.com/opengeospatial/geoparquet/issues/41#issuecomment-1079157251

I also find the empty array a bit less clear in indicating "unknown". Another option could also be to have the field required and allow either the single string "Unknown" or either a list of minimum length 1.

tschaub:
```go
len(geometry_types) == 0
```
vs
```go
geometry_types[0] = "Unknown"
```

(I misspelled Unknown twice while trying to type that :)

m-mohr: Yeah, I'm not a big fan of "Unknown" either. Also, thinking a bit more about STAC it seems for everything we made required to make people think about it, it did not really work that way. People still just push things out without validation, so I could also see that people just omit geometry_types although required. So I tend towards not required and minLength: 1 for the array. As an explicit way to formulate "unknown" we could also consider "null".

tschaub:
> is there ever a case when something like `["Polygon", "Unknown"]` would be useful? Like if the writer knew that some of the geometries were polygons but couldn't guarantee that _all_ geometries were that type?

Oh I hope not :)

My thinking on this is that `geometry_types` is a useful signal for systems that are reading GeoParquet and writing/storing in some representation where it is required to know ahead of time the geometry types that might be read.  It is sensible to include this information in a Parquet file when that file is used as an interchange format because the metadata occurs in the footer and a writer may determine the list of geometry types while writing in a streaming manner.  If it had to be in the header, the writer would have to buffer, and it might seem onerous for us to require it.  If the writer doesn't go to the trouble of determining the geometry types when writing (or somehow can't know it), we still allow them to write GeoParquet, and then the burden is on the reader to buffer before storing/writing if it needs to know all the geometry types.

It's hard for me to imagine where `["Polygon", "Unknown"]` would provide any more information than `[]`.  In either case, if you are a reader that needs to know what kind of geometries you might read before you can store anything, you are going to have to buffer.  Maybe it would be useful if you just wanted to scan a bunch of files and come up with a list of those that probably have polygons?  I'm sure someone can come up with another use case, but I don't think the world will be a worse place if we don't allow it.

rouault:
> Random thought: is there ever a case when something like `["Polygon", "Unknown"]` would be useful? Like if the writer knew that some of the geometries were polygons but couldn't guarantee that _all_ geometries were that type?

I may be the one that initially pushed for unknown. Let me give the context. The idea is that you may have schemas for layers where you want to be able to put any type of geometries. In some files, geometries might be homogenous (let say only points), in other files mixed. Or you may have homogenous individual files, but one is points only and another ones is polygons only. But in the end, you want that the layer schema says "there might be any type of geometry" so that when you import it to something else  (e.g importing a collection of GeoParquet files to a single GeoPackage or PostGIS layer) or compute the union of several files, you get this "any type of geometry" property.

For a writer perspective,
- either the writer knows in advance all geometries it has to write (doing a pre-scan), and then it can write all geometry types
- or it doesn't know that before hand from source metadata and doesn't want to do such a pre-scan, and then it will write [] / unknown

So in that context ``["Polygon", "Unknown"]`` would not seem to be useful.

kylebarron (approved):

tschaub:
@jorisvandenbossche - are you also ok with the latest?

cholmes (approved):

PR #141, opened 2022-11-07
Add basic valid and invalid tests for the json schema

jorisvandenbossche:
Closes https://github.com/opengeospatial/geoparquet/issues/135 (rework of https://github.com/opengeospatial/geoparquet/pull/134 to focus on json schema for now)


tests/validate_schemas.py
```diff
@@ -0,0 +1,57 @@
...
+        valid = False
+        print(f"  - {error.json_path}: {error.message}", "warning")
+        if "description" in error.schema:
+            print(f"    \"{error.schema['description']}\"", "warning")
+
+    if valid:
+        print("This is a valid GeoParquet file.\n", "success")
+    else:
+        print("This is an invalid GeoParquet file.\n", "error")
+        exit(1)
```
jorisvandenbossche, line 33: I started to write this file based on the existing code we had for `geoparquet_validator/__init__.py`, but if just focusing on testing those json files against the schema, it could probably be simplified. Using something like pytest might actually made it easier to have clearer output if it fails.

jorisvandenbossche:
cc @m-mohr do you think this is a sensible approach to test the json schema? (ensure it passes when it should and errors when it should) 
Feedback would be very welcome (I never did this before)

tests/validate_schemas.py
kylebarron: I'm +1 on using pytest

m-mohr:
I'll have a look in the next days :-)

jorisvandenbossche:
OK, I updated this PR for the latest changes in the spec and updated to use pytest.

rouault (approved):

m-mohr:
Looks reasonable to me. Why commit the test files when you can always generate them on the fly though? May lead to PRs where people update them by hand...

cholmes (approved):

jorisvandenbossche:
@m-mohr thanks for those additional cases! This one I don't fully understand what it is testing for:

```
metadata = copy.deepcopy(metadata_template)
metadata["columns"]["primary_geometry"] = ""
invalid_cases["empty_primary_geometry"] = metadata
```

tests/test_json_schema.py
```diff
@@ -0,0 +1,258 @@
...
+metadata.pop("primary_column")
+invalid_cases["missing_primary_column"] = metadata
+
+metadata = copy.deepcopy(metadata_template)
+metadata.pop("columns")
+invalid_cases["missing_columns"] = metadata
+
+metadata = copy.deepcopy(metadata_template)
+metadata["columns"] = {}
+invalid_cases["missing_columns_entry"] = metadata
```
jorisvandenbossche, line 61: This one is still failing (it is xfailed for now to have the tests passing):

```
E           AssertionError: This is an invalid GeoParquet file, but no validation error occurred for 'missing_columns_entry':
E           {
E             "geo": {
E               "columns": {},
E               "primary_column": "geometry",
E               "version": "0.5.0-dev"
E             }
E           }
```

But I suppose that should be easy to solve to require a minimum length of one for `columns`?

m-mohr:
@jorisvandenbossche In #153 we require a minimum length of 1 for the primary geometry and I'm trying to check whether the schema fails for an empty primary geometry.

But I just now realize that it's wrong and should be:
```py
metadata = copy.deepcopy(metadata_template)
metadata["primary_geometry"] = ""
invalid_cases["empty_primary_geometry"] = metadata
```

jorisvandenbossche:
OK, that one is already included!

tests/test_json_schema.py
m-mohr: Ah, I see. We can fix that, sure. Just add `"minProperties": 1,` in the columns schema.
By the way, the written spec doesn't forbid an empty column object.

m-mohr: Example:

```json
    "columns": {
      "type": "object",
      "minProperties": 1,
      "patternProperties": {
        ".+": {

```

tschaub:
Would there be support for moving the `test_json_schema.py` script to the `scripts` directory?  It feels like that directory has the most complete story around managing python dependencies.

jorisvandenbossche:
I wouldn't mind that, certainly now it is only a single file (I am also already making use of the python dependencies that are installed for the scripts to run the tests in CI)

tschaub:
I moved this script in #159 and updated it to read the version constant from the schema file.  So the tests now assert that metadata without a version or with a version that is different than the schema version is invalid.

PR #144, opened 2022-11-18
Allow for more than 3 dimensions in the bbox

tschaub:
The GeoJSON spec recommends 2 or 3 dimensions, but does not forbid geometries that have more.  See https://www.rfc-editor.org/rfc/rfc7946#section-3.1.1.

In addition, the `bbox` length is `2*n` where `n` is the number of dimensions.  See https://www.rfc-editor.org/rfc/rfc7946#section-5.

### DIFF
--- a/format-specs/schema.json
+++ b/format-specs/schema.json
@@ -65,20 +65,9 @@
               "type": "array",
               "description": "Bounding Box of the geometries in the file, formatted according to RFC 7946, section 5.",
               "items": {
-                "type": "number"
-              },
-              "oneOf": [
-                {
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
See my comment at https://github.com/opengeospatial/geoparquet/pull/88#issuecomment-1320479016, we currently only allow 2D or 3D. 

This description was added in https://github.com/opengeospatial/geoparquet/pull/45

tschaub:
Thanks for the clarification, @jorisvandenbossche.

PR #145, opened 2022-11-18
Clarify geographic vs non-geographic bbox values

### LINKED ISSUE #112: bbox crossing the antimeridian
Currently the specification says:

> The bbox, if specified, must be encoded with an array containing the minimum and maximum values of each dimension: `[<xmin>, <ymin>, <xmax>, <ymax>]`. This follows the GeoJSON specification ([RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5)).

"Minimum" and "maximum" are not correct for cases where the bbox crosses the dateline, at least if GeoJSON is used as the basis. See GeoJSON 5.2, the following is a rough bbox for the Fiji archipelago, spanning 5 degrees of longitude: `[177.0, -20.0, -178.0, -16.0]`.

If GeoJSON is followed, the wording should be updated/clarified. If it should be strictly minimum/maximum, then the GeoJSON reference should be removed. 

Happy to create a PR for review, but it would help to understand the intention for GeoParquet. I personally would prefer the first option and be consistent with GeoJSON, which is also what is used in OGC API Features and I think STAC.

tschaub:
This updates the language about bbox to align with GeoJSON for geographic coordinates and give min/max examples for non-geographic coordinates.

Fixes #112.

(One thing to note - the GeoJSON spec is a bit misleading in saying that the bbox has values for _"all axes of the most southwesterly point followed by all axes of the more northeasterly point."_  What it means is that the value for $i_0$ is the most westerly, the value for $i_1$ is the most southerly, the value for $i_n$ is the most easterly, and the value for $i_{n+1}$ is the most northerly.  The values for all other items are the minima followed by the maxima.  The way the spec reads, it sounds like the third item of a bbox for geometries with a z would be the z value for the most southwesterly point.  Instead it should be the minimum z value for all coordinates.  I'll bring this up on the GeoJSON list and submit an errata report if others agree.)

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -168,11 +168,18 @@ Bounding boxes are used to help define the spatial extent of each geometry colum
 Implementations of this schema may choose to use those bounding boxes to filter
 partitions (files) of a partitioned dataset.
 
-The bbox, if specified, must be encoded with an array containing the minimum
-and maximum values of each dimension: `[<xmin>, <ymin>, <xmax>, <ymax>]` for 2D
-or `[<xmin>, <ymin>, <zmin>, <xmax>, <ymax>, <zmax>]` for 3D geometries.
-This follows the GeoJSON specification ([RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5)),
-but is expressed in the same CRS as the geometry.
+The bbox, if specified, must be encoded with an array representing the range of values for each dimension in the
+geometry coordinates.  For geometries in a geographic coordinate reference system, longitude and latitude values are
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
> (One thing to note - the GeoJSON spec is a bit misleading in saying that the bbox has values for _"all axes of the most southwesterly point followed by all axes of the more northeasterly point."_ ...

Yes, agreed with your analysis. I assume that for the z dimension it is actually min/max, and not the z of the most southwesterly / northeasterly points. But that's indeed not clear from the GeoJSON description.

jorisvandenbossche (commented):
Looks good!

jorisvandenbossche (approved):

PR #149, opened 2022-11-22
More consistent spelling and punctuation for JSON types

### LINKED ISSUE #128: crs data type: JSON object

tschaub:
This change makes more consistent use of the JSON types "object" (instead of JSON object or Map) and "number" (instead of double).  Where we discuss string literals in the JSON metadata, they are double quoted and formatted as code (e.g. `"like this"`).

Fixes #128.


m-mohr (approved):
I'm not a big fan of quotes around the strings in documentation, but otherwise looks good to me.

kylebarron (approved):

jorisvandenbossche (approved):

PR #151, opened 2022-12-04
schema.json: update reference to projjson schema to v0.5

rouault:
Latest released is v0.5 (cf https://proj.org/specifications/projjson.html). It is fully backward compatible with previous versions.


kylebarron (approved):

m-mohr (approved):
Great, yet another thing that we don't need to cater for in #131 then.

PR #153, opened 2022-12-05
Require minimum length of 1 for primary_geometry #129

m-mohr:
Require a minimum length of 1 for the primary_geometry as otherwise it can be empty.
+ some minor test/formatting changes.

Partially solves #129.

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -38,7 +38,7 @@ All file-level metadata should be included under the "geo" key in the parquet me
 |     Field Name     |  Type  |                             Description                              |
 | ------------------ | ------ | -------------------------------------------------------------------- |
 | version     		 | string | **REQUIRED** The version of the GeoParquet metadata standard used when writing. |
-| primary_column     | string | **REQUIRED** The name of the "primary" geometry column with a minimum length of 1. |
+| primary_column     | string | **REQUIRED** The name of the "primary" geometry column. |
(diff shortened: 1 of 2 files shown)


format-specs/geoparquet.md
```diff
@@ -37,9 +37,9 @@ All file-level metadata should be included under the "geo" key in the parquet me
 
 |     Field Name     |  Type  |                             Description                              |
 | ------------------ | ------ | -------------------------------------------------------------------- |
-| version     		 | string | **REQUIRED** The version of the GeoParquet metadata standard used when writing.   |
-| primary_column     | string | **REQUIRED** The name of the "primary" geometry column.                |
-| columns            | object<key, [Column Metadata](#column-metadata)> | **REQUIRED** Metadata about geometry columns, with each key is the name of a geometry column in the table. |
+| version     		 | string | **REQUIRED** The version of the GeoParquet metadata standard used when writing. |
+| primary_column     | string | **REQUIRED** The name of the "primary" geometry column with a minimum length of 1. |
```
jorisvandenbossche, line 41: I don't know if it's necessarily need to mention it here as well (I find it somewhat obvious that a column name is not an empty string), but either way is fine.

jorisvandenbossche (approved):
Thanks, looks good!

format-specs/geoparquet.md
cholmes:
```suggestion
| primary_column     | string | **REQUIRED** The name of the "primary" geometry column |
```

PR #154, opened 2022-12-05
The default value of the crs field is required #152

### LINKED ISSUE #152: crs fields it too vague
The CRS field states (bold text by me):

> If the crs field is not included then the data in this column must be stored in longitude, latitude based on the WGS84 datum, and CRS-aware implementations **should assume** a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84)

This is weird. Is it the default value or not? Stating that this **might** be the default value is not very helpful. 

So I propose to rephrase to something like:

> The default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means the data in this column must be stored in longitude, latitude based on the WGS84 datum.

m-mohr:
Fixes #152

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -62,7 +62,7 @@ Each geometry column in the dataset must be included in the columns field above 
 | --- | --- | --- |
 | encoding | string | **REQUIRED** Name of the geometry encoding format. Currently only `"WKB"` is supported. |
 | geometry_types | \[string] | **REQUIRED** The geometry types of all geometries, or an empty array if they are not known. |
-| crs | object | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) object representing the Coordinate Reference System (CRS) of the geometry. If the crs field is not included then the data in this column must be stored in longitude, latitude based on the WGS84 datum, and CRS-aware implementations should assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84). |
+| crs | object | **OPTIONAL** [PROJJSON](https://proj.org/specifications/projjson.html) object representing the Coordinate Reference System (CRS) of the geometry. If the field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means the data in this column must be stored in longitude, latitude based on the WGS84 datum. |
 | orientation | string | **OPTIONAL** Winding order of exterior ring of polygons. If present must be `"counterclockwise"`; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
 | edges | string | **OPTIONAL** Name of the coordinate system for the edges. Must be one of `"planar"` or `"spherical"`. The default value is `"planar"`. |
 | bbox | \[number] | **OPTIONAL** Bounding Box of the geometries in the file, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |


tschaub (approved):

format-specs/geoparquet.md
```diff
@@ -80,10 +80,10 @@ which itself implements the model of
 Apart from the difference of encodings, the semantics are intended to match
 WKT2:2019, and a CRS in one encoding can generally be represented in the other.
 
-If CRS is not provided, all coordinates in the geometry must use longitude, latitude
-based on the WGS84 datum to store their data.
+If the `crs` field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84),
+which means that all coordinates in the geometries must use longitude, latitude based on the WGS84 datum.
 
-If an implementation is CRS-aware and needs a CRS representation of the data it should assume a default value is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which is equivalent to the well-known [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) but changes the axis from latitude-longitude to longitude-latitude.
+[OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84) is equivalent to the well-known [EPSG:4326](https://epsg.org/crs_4326/WGS-84.html) but changes the axis from latitude-longitude to longitude-latitude.
```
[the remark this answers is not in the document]
cholmes: No issue. That twisted language came from some idea that there is data that people put in lat/long that is not exactly crs84, due to some sub-parameters of the WKT, so this was to somehow communicate that it doesn't need to be exact, or something... I forget the exact thing that lead to this. But since I can barely remember what we were trying to achieve I'm in support of clearer language.

format-specs/geoparquet.md
```diff
@@ -80,10 +80,10 @@ which itself implements the model of
 Apart from the difference of encodings, the semantics are intended to match
 WKT2:2019, and a CRS in one encoding can generally be represented in the other.
 
-If CRS is not provided, all coordinates in the geometry must use longitude, latitude
-based on the WGS84 datum to store their data.
+If the `crs` field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84),
```
cholmes, line 83:
```suggestion
If the `crs` field is not provided, all coordinates in the geometries must use longitude, latitude based on the WGS84 datum,
```

format-specs/geoparquet.md
```diff
@@ -80,10 +80,10 @@ which itself implements the model of
...
+which means that all coordinates in the geometries must use longitude, latitude based on the WGS84 datum.
```
cholmes, line 84:
```suggestion
which means CRS-aware implementations can assume a default value of [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84)
```

I think switching the order as suggested in these two lines can get across that if you are implementing and don't know about CRS's then you can just use long/lat. I do think it'd be good to not make it seem like to implement geoparquet you need a full projection engine - you could just read and write only long/lat. 

If we were to do a full OGC-style spec we'd likely have the core conformance just be long/lat, and another conformance class add in more projections (crs field and ability to understand it)

cholmes (approved):
Looks good. Made a suggestion that hopefully clears up the language some but preserves a bit of the original intent.

format-specs/geoparquet.md
m-mohr: We tried to avoid the "can assume" wording as this is weird, so another proposal:

```suggestion
which means the default value is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84) for CRS-aware implementations
```

format-specs/geoparquet.md
m-mohr: all coordinates in the geometries must use longitude, latitude based on the WGS84 datum

m-mohr:
```suggestion
all coordinates in the geometries must use longitude, latitude based on the WGS84 datum
```

format-specs/geoparquet.md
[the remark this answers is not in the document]
m-mohr: Could you have a look again? I tried to squeeze out the "can assume" while trying to keep your intention.

format-specs/geoparquet.md
```diff
@@ -80,10 +80,10 @@ which itself implements the model of
 Apart from the difference of encodings, the semantics are intended to match
 WKT2:2019, and a CRS in one encoding can generally be represented in the other.
 
-If CRS is not provided, all coordinates in the geometry must use longitude, latitude
-based on the WGS84 datum to store their data.
+all coordinates in the geometries must use longitude, latitude based on the WGS84 datum
```
m-mohr, line 83:
```suggestion
If CRS is not provided, all coordinates in the geometries must use longitude, latitude based on the WGS84 datum
```

format-specs/geoparquet.md
```diff
@@ -80,10 +80,10 @@ which itself implements the model of
 Apart from the difference of encodings, the semantics are intended to match
 WKT2:2019, and a CRS in one encoding can generally be represented in the other.
 
-If CRS is not provided, all coordinates in the geometry must use longitude, latitude
-based on the WGS84 datum to store their data.
+If CRS is not provided, all coordinates in the geometries must use longitude, latitude based on the WGS84 datum
```
tschaub, line 83:
```suggestion
If CRS is not provided, all coordinates in the geometries must use longitude, latitude based on the WGS84 datum,
```

tschaub:
@m-mohr - looks like this will need a rebase after #156.

m-mohr:
@tschaub Indeed, done.

tschaub (approved):

PR #156, opened 2022-12-07
Clean-up README

m-mohr:
Some proposals for a cleaner README:
- Directly link to examples and JSON Schema directly so that they are easy to find (the schema wasn't linked to at all from the spec, the spec itself wasn't linked to at all from the repo README)
- Update version numbers to one consistent value instead of 0.4.0 and 0.5.0-dev in different places
- Markdown tweaks, e.g. code fences
- Remove "Optional" as it is in STAC to make it easier to see the difference directly. Only required fields are declared, all other are implicitly optional.
- Don't hide links somewhere in the text

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -2,12 +2,16 @@
 
 ## Overview
 
-The [Apache Parquet][parquet] provides a standardized open-source columnar storage format. This specification defines how geospatial data
+The [Apache Parquet](https://parquet.apache.org/) provides a standardized open-source columnar storage format. This specification defines how geospatial data
 should be stored in parquet format, including the representation of geometries and the required additional metadata.
+
+**Additional resources:**
+* [Examples](../examples/)
+* [JSON Schema](schema.json)
 
 ## Version
 
-This is version 0.4.0 of the GeoParquet specification.
+This is version 0.5.0-dev of the GeoParquet specification.
 
 ## Geometry columns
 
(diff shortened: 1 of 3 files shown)


format-specs/geoparquet.md
```diff
@@ -52,21 +56,21 @@ but have a default geometry used for geospatial operations.
 
 #### version
 
-Version of the GeoParquet spec used, currently 0.4.0
+Version of the GeoParquet spec used, currently 0.5.0-dev
```
jorisvandenbossche, line 59: Something I am wondering about this version number. While it is consistent to use this version here (as that is also used elsewhere, in the schema.json, etc), in practice we also don't really want that anyone would actually create files with this version? 
Do we expect readers to understand that version? (or error for it?)

I am wondering that for readers of the spec, it might be more useful to say here "one of the released versions" instead of actually "currently 0.5.0-dev" ?

tschaub: @jorisvandenbossche - I agree we need to do more cleanup on this.  I think it would be enough for the description of the `version` metadata to say "The version identifier for the GeoParquet specification."  Or we could include language about the "release" version as well.  I don't think we need to repeat the actual version in this description.  In addition, I think the whole "Additional file metadata information" section could be deleted, with the detail instead going in the description column in the table above.

I think the changes in this branch are an incremental improvement, so I'll merge as is.  And then let's keep iterating on improvements.

jorisvandenbossche: Yes, just removing the version number altogether in this location sounds good

tschaub: Since #159 is about limiting the number of places the version identifier is repeated, I've made this change there.

Issue #146, opened 2022-11-21
Propose and document release process


cholmes:
We should have a clear cookbook of how we cut a new release.

tschaub:
I am hoping we can get in #159 before doing this.  That change removes 4 places that the version identifier is repeated and merging it will help simplify the release process.

tschaub:
I've added a wiki page with the [release process](https://github.com/opengeospatial/geoparquet/wiki/Release-Process).  I followed this process in #161, and here is the resulting release: [v1.0.0-beta.1](https://github.com/opengeospatial/geoparquet/releases/tag/v1.0.0-beta.1).

PR #159, opened 2022-12-07
Read version number from the schema

tschaub:
The goal of this branch is to reduce the number of places the version number is repeated.  Here is a summary of the changes and motivation for each:

 * The `examples/example.py` script currently includes the version number.  I changed this to read from `format-specs/schema.json` instead.  The `examples` directory has an `environment.yml` file that apparently allows `examples/example.py` to be run, but because there was no readme, I moved the script to the `scripts` directory where there are instructions on how to install python dependencies.
 * The `tests/test_json_schema.py` script currently includes the version number.  As above, I changed this to read from `format-specs/schema.json` instead.  The script also has dependencies and no instructions on how to install them, so I moved it to the `scripts` directory where this is documented and updated the dependency list there.
 * I updated the `.github/workflows/scripts.yml` workflow to run the script that generates the example data before asserting that the example metadata matches expectations.  This should catch cases where someone updates the expected metadata but not the example or vice versa (if someone forgets to change both but changes the metadata schema, ideally the validator job should catch that).
 * I removed one additional place where the version identifier was repeated in the spec.


kylebarron:
Looks like https://github.com/opengeospatial/geoparquet/pull/24 was what added `environment.yml` (cc @TomAugspurger). I personally prefer poetry and pip/PyPI for packages, but maybe on Windows Conda is still easier.

tschaub:
One issue with the script that generates the `example.parquet` is that it depends on whatever `geopandas.datasets.get_path("naturalearth_lowres")` resolves to.  I've seen that the resulting parquet dataset can vary.  For example, the "crs" data may have a `"$schema": "https://proj.org/schemas/v0.5/projjson.schema.json"` or `"$schema": "https://proj.org/schemas/v0.4/projjson.schema.json"`.  I wonder if this depends on the specific version of geopandas.  If so, can Poetry help us get something more repeatable here?

kylebarron:
> Perhaps we don't expect a bitwise match for parquet files generated on different systems.

Parquet writers usually include the name of the implementation in the metadata, so it won't be bitwise equal across implementations

```py
import pandas as pd
import pyarrow.parquet as pq
df = pd.DataFrame({'a': [1, 2, 3, 4]})
df.to_parquet('tmp.parquet')
pq.read_metadata('tmp.parquet').created_by
# 'parquet-cpp-arrow version 10.0.1'
```

jorisvandenbossche:
> One issue with the script that generates the `example.parquet` is that it depends on whatever `geopandas.datasets.get_path("naturalearth_lowres")` resolves t

It's probably good to rely on this for the example data. For one, we don't guarantee it to be stable as we might update it with new naturalearth data releases (which is maybe not a problem for our use case), but the dataset also has some political problems (like Crimea belonging to Russia, which is a reason we would prefer to remove it from geopandas on the long term).

Maybe we can find some other online data source to use as example?

jorisvandenbossche:
I personally prefer conda, and would rather not use poetry. But I also don't think it is worth supporting both since in the end it is just for running this script, which is not something one will have to do often yourself. So I am OK with continue to use poetry here. 

But we should maybe avoid relying on GDAL (fiona) then to read in the example data to keep the installation of the env for the scripts simpler (eg if we read from a geojson file, GDAL is not necessarily needed).

tschaub:
> But we should maybe avoid relying on GDAL (fiona) then to read in the example data to keep the installation of the env for the scripts simpler (eg if we read from a geojson file, GDAL is not necessarily needed).

I agree that we could reduce dependencies and simplify creation of the example file.  And I agree that something besides political boundaries would be good.  Let's take this up separately from the changes here.

scripts/pyproject.toml
```diff
@@ -13,10 +13,10 @@ pygeos = "^0.12.0"
 pandas = "^1.4.2"
 click = "^8.1.2"
 
-[tool.poetry.dev-dependencies]
-ipykernel = "^6.13.0"
-isort = "^5.10.1"
-black = "^22.3.0"
+[tool.poetry.group.dev.dependencies]
```
kylebarron, line 16: Oh interesting; this must be a change with poetry 1.2 from `dev-dependencies` to `group.dev`

scripts/pyproject.toml
```diff
@@ -13,10 +13,10 @@ pygeos = "^0.12.0"
 pandas = "^1.4.2"
 click = "^8.1.2"
 
-[tool.poetry.dev-dependencies]
-ipykernel = "^6.13.0"
-isort = "^5.10.1"
```
kylebarron, line 18: ipykernel is fine to remove; I tend to use that myself for interactive debugging of scripts. I'd have a preference for keeping isort though, because then (along with black) we have a deterministic formatting of any scripts, no matter who's editing it

scripts/pyproject.toml
tschaub: Yeah, if we need to support poetry < 1.2, I could move these back.  `poetry add --dev ...` (which has been replaced with `poetry add --group dev`) now writes to this group instead of `dev-dependencies`.  See https://python-poetry.org/docs/managing-dependencies/#dependency-groups for some additional detail.

kylebarron: I'm fine with keeping >=1.2

scripts/pyproject.toml
tschaub: I added `isort` back in the latest commit.

I think it is nice to work on a repo where the dependencies are used (ideally in a CI job).  So for example, if we were running `black --check` in a CI job, that would justify the `black` dependency.  And we could have a job that exercised `isort` as well.  (FWIW, I think doing both of these things would be great - I love not thinking about whitespace or formatting and having tools take care of that for me.)

kylebarron: Would you rather have those lint steps in a CI job or in the pre commit config? Or should we remove the pre commit config altogether?

tschaub: CI job

tschaub: This is probably controversial, but I find pre-commit hooks developer hostile.  I like to use commits as a place to persist work in progress.  Often that work in progress won't pass all tests (linter rules, etc).  Having a pre-commit hook that makes it harder to push work in progress commits feels like a hassle.  I think there is a ton of value in these rules, tests, etc.  CI jobs that prevent merging feels like the appropriate place to enforce them.

jorisvandenbossche: Yes `--no-verify` is your friend ;) (`-n` for short)

While I often use --no-verify for exactly the reason you describe above, I also love pre-commit because it makes me to have to think so much less about linting in general.

Also, you don't _have_ to install the pre-commit hooks in a repo. So even if we have `.pre-commit-config.yaml` and use that also in CI to run the linting checks, you are not obliged to use that locally.

tschaub: Thanks, I do know how to disable/ignore them - just saying that I think in general a pre-commit hook makes it harder to push WIP commits.  I haven't done anything in this branch related to the pre-commit hooks, but good discussion.

tschaub:
@kylebarron or @jorisvandenbossche - Any more comments on these changes?  Definitely have more to discuss and probably do related to CI, dependencies, example data, etc.  But I'm wondering if we can get in the changes that reduce the duplicated version identifiers to simplify the release process (#146).

jorisvandenbossche (approved):
Yes, all my suggestions about a different data set / not using GDAL were just ideas that can certainly be left for other issues/PRs

PR #160, opened 2022-12-13
Refer to RFC 2119 for definition of requirement levels

### LINKED ISSUE #26: Use RFC 2119 definitions of MUST, SHOULD, etc.?

tschaub:
This adds a link to [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) for the definition of "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL".

I changed a number of instances of "must" to "MUST" (etc.) to conform with convention.  I didn't change all.  Personally I feel like this is enough to unblock the 1.0.0-beta.1 release and that we can keep iterating on formalizing the requirements (but I'm also open to suggested changes).

Fixes #26.

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -9,22 +9,24 @@ should be stored in parquet format, including the representation of geometries a
 * [Examples](../examples/)
 * [JSON Schema](schema.json)
 
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED",  "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt).
+
 ## Version
 
 This is version 0.5.0-dev of the GeoParquet specification.
 
 ## Geometry columns
 
-Geometry columns are stored using the `BYTE_ARRAY` parquet type. They are encoded as [WKB](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary).
+Geometry columns MUST be stored using the `BYTE_ARRAY` parquet type. They MUST be encoded as [WKB](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary).
 See the [encoding](#encoding) section below for more details.
 
 ### Nesting
 
-Geometry columns must be at the root of the schema. A geometry cannot be a group field or nested in a group. In practice, this means that when writing to GeoParquet from another format, geometries cannot be contained in complex or nested types such as structs, lists, arrays, or map types.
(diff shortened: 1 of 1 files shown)


cholmes (approved):
Looks good, thanks for doing this!

format-specs/geoparquet.md
```diff
@@ -112,7 +112,7 @@ column (and not per geometry).
...
-of the spec may support alternative encodings. This should be the ["OpenGIS® Implementation Specification for Geographic information - Simple feature access - Part 1: Common architecture"](https://portal.ogc.org/files/?artifact_id=18241) WKB representation (using codes for 3D geometry types in the \[1001,1007\] range). This encoding is also consistent with the one defined in the ["ISO/IEC 13249-3:2016 (Information technology - Database languages - SQL multimedia and application packages - Part 3: Spatial)"](https://www.iso.org/standard/60343.html) standard.
+of the spec may support alternative encodings. This SHOULD be the ["OpenGIS® Implementation Specification for Geographic information - Simple feature access - Part 1: Common architecture"](https://portal.ogc.org/files/?artifact_id=18241) WKB representation (using codes for 3D geometry types in the \[1001,1007\] range). This encoding is also consistent with the one defined in the ["ISO/IEC 13249-3:2016 (Information technology - Database languages - SQL multimedia and application packages - Part 3: Spatial)"](https://www.iso.org/standard/60343.html) standard.
```
jorisvandenbossche, line 115: _Should_ this be a MUST?

tschaub: Probably.  I was wondering the same and should have asked.  Thought maybe there had been previous justification for making it a requirement instead of a recommendation.

jorisvandenbossche: Yeah, to be honest I am not sure how important this is. I only know well the GEOS WKB parser, and that one is flexible in parsing both ISO and extended flavors out of the box, so the geopandas reader for geoparquet wouldn't actually care what flavor of WKB is used. 

(and we actually also need to update our writer to use the correct flavor to honor the spec!)

PR #172, opened 2023-06-07
Parquet Geospatial Compatibility Recommendations

cholmes:
This is still a work in progress, but wanted to get it up to elicit feedback.

Big questions:

* Does the naming sound right? Are there alternatives?
* Does it successfully walk the line of explaining this option but not making it a 'spec' where people would only implement it and say they make 'GeoParquet'.

I put it in 'format-specs', but could put it elsewhere or rename that directory.

Any other tweaks / feedback are welcome as well.

### DIFF
--- a/format-specs/compatible-parquet.md
+++ b/format-specs/compatible-parquet.md
@@ -1,0 +1,42 @@
+
+# Parquet Geospatial Compatibility
+
+The goal of GeoParquet is that every tool producing Parquet and includes geospatial data uses our [defined metadata](./geoparquet_spec) to enable maximum compatibility. But in the interim time when only some tools properly produce GeoParquet, we are recommending the following best practices for the distribution of geospatial data as Parquet. 
+
+This is *only* recommended for those who are using tools that don't yet produce valid GeoParquet, and we encourage advocating to the creaters of tools you are using to implement the GeoParquet spec. Feel free to [start a discussion](https://github.com/opengeospatial/geoparquet/discussions) to raise awareness of tools that ideally support Geoparquet - the community can likely help in encouraging an implementation.
+
+Tools and libraries that read GeoParquet are encouraged to be able to parse these files, to make it easy to get data into the GeoParquet ecosystem. But it is only recommended for tools and libraries to produce valid GeoParquet, following [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle) of being liberal in what you accept but conservative in what you send.
+
+## Compatibility Guidelines
+
+TODO: flesh this out with more narrative.
+
+The geometry column should be named 'geometry'.
+
+The geometry column should be either a binary column with Well Known Binary (WKB) or a text column with Well Known Text (WKT) used to define the geometries.
+
+All data is stored in longitude, latitude based on the WGS84 datum.
+
+Readers should be default make the following assumptions when they read one of these columns:
+
+* The geometry_types values is an empty array - the types are not known.
+* The CRS OGC:CRS84
+* No assertions are made on the winding order
+* Edges are planer
+
+## Data Reader Implementation Considerations
+
+Reading this non-compliant geospatial data from Parquet should ideally work with no user intervention if the producer followed all the guidelines and named their geometry column 'geometry'. Readers can optionally support user input (in whatever manner works for the tool / library) to provide hints for metadata that is not inline with the lowest common denominator compatibility. This would include things like letting the user supply a geometry column name (ie something other than 'geometry'), providing a more specific geometry_type value, or providing other enhanced metadata (specifying the CRS, the winding order, the edges, the bbox or the epoch).
+
+We strongly advise against creating a reader that can only understand these geospatial compatible types, all readers should start by looking at the metadata specified by GeoParquet and only fall back on these compatibility techniques if the metadata is not present. A reader that could not read GeoParquet but would read compatible geodata would have no idea if there was in fact metadata, and thus could easily decrease interoperability.
+
+## Data Producer Considerations
+
+As mentioned above, we strongly recommend trying to find tools that will produce valid GeoParquet. If the tool you are working with does not support it directdly then there are [many tools](geoparquet.org) that can help you. Only if there is no way to create valid GeoParquet metadata do we recommend this route. This will enable those who have readers that understand GeoParquet and these compatible files to turn it into valid GeoParquet themselves.
+
+We recommend sticking to the core recommendations as much as possible - naming the geometry column 'geometry', using WKB, and storing data
+as long, lat. If your data must be formatted differently then less readers will be able to work with it. If you do go that route be sure to
+make it clear in all your documentation where things are different.
+
+
+


format-specs/compatible-parquet.md
```diff
@@ -0,0 +1,42 @@
...
+
+Tools and libraries that read GeoParquet are encouraged to be able to parse these files, to make it easy to get data into the GeoParquet ecosystem. But it is only recommended for tools and libraries to produce valid GeoParquet, following [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle) of being liberal in what you accept but conservative in what you send.
+
+## Compatibility Guidelines
+
+TODO: flesh this out with more narrative.
+
+The geometry column should be named 'geometry'.
+
+The geometry column should be either a binary column with Well Known Binary (WKB) or a text column with Well Known Text (WKT) used to define the geometries.
```
rouault, line 16: I believe this was raised in a past discussion, but it is a bit paradoxical to allow WKT geometries in a "compatible GeoParquet" file, whereas the GeoParquet itself only supports WKB.

cholmes: I think if we were calling this thing like 'simple geoparquet' or something with it in the name I'd agree we should put it in the spec. I see this as just a set of recommendations.

But I'd be up to call it WKB in this 'core' description, and then mention WKT as an option to support.

cholmes: Moved WKT to not be in the 'core' description, but an option that readers might add.

format-specs/compatible-parquet.md
```diff
@@ -1,42 +1,45 @@
...
-* Edges are planer
+The above are the key recommendations a data producer should follow. Any implemented reader will need to make the following assumptions when reading one of these columns, unless the user supplies additional information that they are aware of:
+
+* The geometry_types values is an empty array, signaling the geometry type is not known and the reader should make no assumptions about the types, as defined in the [geometry_types](./geoparquet.md#geometry_types) section of the spec. 
+
+* Any CRS-aware reader should assume that the CRS is OGC:CRS84 as explained in the [crs](./geoparquet.md#crs) section of the spec. (Or it could assume it is EPSG:4326 but overriding the axis order to assume longitude latitude as explained in the [Coordinate axis order](./geoparquet.md#coordinate-axis-order)) section.
+
+* No assertions are made on the winding order, the default of the [orientation](./geoparquet.md#orientation) section of the spec.
+
+* Edges are planer, as explained to be the default in the [edges](./geoparquet.md#edges) section of the spec.
```
kylebarron, line 29:
```suggestion
* Edges are planar, as explained to be the default in the [edges](./geoparquet.md#edges) section of the spec.
```

format-specs/compatible-parquet.md
```diff
@@ -1,42 +1,45 @@
 
 # Parquet Geospatial Compatibility
 
-The goal of GeoParquet is that every tool producing Parquet and includes geospatial data uses our [defined metadata](./geoparquet_spec) to enable maximum compatibility. But in the interim time when only some tools properly produce GeoParquet, we are recommending the following best practices for the distribution of geospatial data as Parquet. 
+The goal of GeoParquet is that every tool producing Parquet and includes geospatial data uses official [metadata defined in the GeoParquet spec](./geoparquet_spec) to achieve true interoperability. Thid document represents a set of guidelines for those would like to produce geospatial Parquet data but are using tools that are not yet fully implementing GeoParquet metadata. It is meant to be used just for the interim time when only some tools properly produce GeoParquet, to data producer support the growing ecosystem.
```
kylebarron, line 3:
```suggestion
The goal of GeoParquet is that every tool producing Parquet and includes geospatial data uses official [metadata defined in the GeoParquet spec](./geoparquet_spec) to achieve true interoperability. This document represents a set of guidelines for those would like to produce geospatial Parquet data but are using tools that are not yet fully implementing GeoParquet metadata. It is meant to be used just for the interim time when only some tools properly produce GeoParquet, to data producer support the growing ecosystem.
```

format-specs/compatible-parquet.md
```diff
@@ -0,0 +1,41 @@
...
+
+We strongly advise against creating a reader that can only understand these geospatial compatible files - all readers should start by looking at the metadata specified by GeoParquet and only fall back on these compatibility techniques if the metadata is not present. A reader that could not read GeoParquet but would read compatible geodata would have no idea if there was in fact metadata, and thus could easily decrease interoperability.
+
+## Data Producer Considerations
+
+As mentioned above, we strongly recommend trying to find tools that will produce valid GeoParquet. If the tool you are working with does not support it directdly then there are [many tools](geoparquet.org) that can help you. Only if there is no way to create valid GeoParquet metadata do we recommend this route. This will enable those who have readers that understand GeoParquet and these compatible files to turn it into valid GeoParquet themselves.
```
mojodna, line 39:
```suggestion
As mentioned above, we strongly recommend trying to find tools that will produce valid GeoParquet. If the tool you are working with does not support it directly then there are [many tools](https://geoparquet.org) that can help you. We only recommend this route if if there is no way to create valid GeoParquet metadata. This will enable those who have readers that understand GeoParquet and these compatible files to turn it into valid GeoParquet themselves.
```

format-specs/compatible-parquet.md
```diff
@@ -0,0 +1,41 @@
+# Parquet Geospatial Compatibility
+
+The goal of GeoParquet is that every tool producing Parquet and includes geospatial data uses official [metadata defined in the GeoParquet spec](./geoparquet_spec) to achieve true interoperability. This document represents a set of guidelines for those would like to produce geospatial Parquet data but are using tools that are not yet fully implementing GeoParquet metadata. It is meant to be used just for the interim time when only some tools properly produce GeoParquet, to data producer support the growing ecosystem.
+
+To be clear, this is *only* recommended for those who are using tools that don't yet produce valid GeoParquet, and we encourage advocating to the creaters of tools you are using to implement the GeoParquet spec. Feel free to [start a discussion](https://github.com/opengeospatial/geoparquet/discussions) to raise awareness of tools that ideally support Geoparquet - the community can likely help in encouraging an implementation.
```
mojodna, line 5:
```suggestion
To be clear, this is *only* recommended for those who are using tools that don't yet produce valid GeoParquet, and we encourage advocating to the creaters of tools you are using to implement the GeoParquet spec. Feel free to [start a discussion](https://github.com/opengeospatial/geoparquet/discussions) to raise awareness of tools that ideally support GeoParquet - the community can likely help in encouraging an implementation.
```

format-specs/compatible-parquet.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Data Producer Considerations
+
+As mentioned above, we strongly recommend trying to find tools that will produce valid GeoParquet. If the tool you are working with does not support it directdly then there are [many tools](geoparquet.org) that can help you. Only if there is no way to create valid GeoParquet metadata do we recommend this route. This will enable those who have readers that understand GeoParquet and these compatible files to turn it into valid GeoParquet themselves.
+
+We recommend sticking to the core recommendations as much as possible - naming the geometry column 'geometry', using WKB, and storing data as long, lat. If your data must be formatted differently then less readers will be able to work with it. If you do go that route be sure to make it clear in all your documentation where things are different.
```
mojodna, line 41:
```suggestion
We recommend sticking to the core recommendations as much as possible - naming the geometry column 'geometry', using WKB, and storing data as long, lat. If your data must be formatted differently, fewer readers will be able to work with it. If you do go that route, be sure to make it clear in all your documentation where things are different.
```

format-specs/compatible-parquet.md
```diff
@@ -0,0 +1,41 @@
...
+
+* No assertions are made on the winding order, the default of the [orientation](./geoparquet.md#orientation) section of the spec.
+
+* Edges are planar, as explained to be the default in the [edges](./geoparquet.md#edges) section of the spec.
+
+## Data Reader Implementation Considerations
+
+Reading this non-compliant geospatial data from Parquet should ideally work with no user intervention if the producer followed all the guidelines and named their geometry column 'geometry'. Readers can optionally support user input (in whatever manner works for the tool / library) to provide hints for metadata that is not inline with the lowest common denominator compatibility. This would include things like letting the user supply a geometry column name (ie something other than 'geometry'), using Well Known Text (WKT) in a `STRING` column instead of WKB, providing a more specific geometry_type value, or providing other enhanced metadata (specifying the CRS, the winding order, the edges, the bbox or the epoch).
```
mojodna, line 33: Assuming that `STRING` geometry columns contain WKT is a clever fallback.

mojodna (commented):
Thanks for floating this. It's a nice compromise for those of us who don't control the development/deployment cycle of the readers/writers we use.

format-specs/compatible-parquet.md
jorisvandenbossche: This came up in the WKT vs WKB discussion as well, and I think if we want to allow WKT here in the "compatible" files, we should just officially add it as a possible encoding in the spec (https://github.com/opengeospatial/geoparquet/discussions/170#discussioncomment-5817223)

format-specs/compatible-parquet.md
```diff
@@ -0,0 +1,41 @@
...
+
+The above are the key recommendations a data producer should follow. Any implemented reader will need to make the following assumptions when reading one of these columns, unless the user supplies additional information that they are aware of:
+
+* The geometry_types values is an empty array, signaling the geometry type is not known and the reader should make no assumptions about the types, as defined in the [geometry_types](./geoparquet.md#geometry_types) section of the spec.
+
+* Any CRS-aware reader should assume that the CRS is OGC:CRS84 as explained in the [crs](./geoparquet.md#crs) section of the spec. (Or it could assume it is EPSG:4326 but overriding the axis order to assume longitude latitude as explained in the [Coordinate axis order](./geoparquet.md#coordinate-axis-order)) section.
+
+* No assertions are made on the winding order, the default of the [orientation](./geoparquet.md#orientation) section of the spec.
+
+* Edges are planar, as explained to be the default in the [edges](./geoparquet.md#edges) section of the spec.
```
jorisvandenbossche, line 29: I am wondering how practical this requirement / assumption is. 
Assume you are using BigQuery, and you have a geography column, and you then export the table to Parquet (https://cloud.google.com/bigquery/docs/exporting-data#parquet_export_details). The docs don't mention how it will represent a geography column, but I assume it will be WKT or WKB. And I also assume that it will then use spherical edges, since this is what Bigquery assumes when loading data from WKT/WKB (https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_geogfromtext). And I don't directly see a way to manually create a new column of binary/string type with the WKB/WKT values that have planar edges (the `ST_ASTEXT` or `ST_ASBINARY` don't provide any options for it)

(of course, we could say this will require the user to specify the edges through some option when reading this kind of data. But if you didn't produce the data yourself, you might not necessarily know this)

format-specs/compatible-parquet.md
```diff
@@ -0,0 +1,41 @@
+# Parquet Geospatial Compatibility
+
+The goal of GeoParquet is that every tool producing Parquet and includes geospatial data uses official [metadata defined in the GeoParquet spec](./geoparquet_spec) to achieve true interoperability. This document represents a set of guidelines for those would like to produce geospatial Parquet data but are using tools that are not yet fully implementing GeoParquet metadata. It is meant to be used just for the interim time when only some tools properly produce GeoParquet, to data producer support the growing ecosystem.
```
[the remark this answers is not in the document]
cholmes: Cool, will fix - not sure what I was thinking.

format-specs/compatible-parquet.md
cholmes: I agree it's not a great assumption, but I think we have to assume something? Unless there's some way to automatically detect this? 

Hmmm, just had an idea that might be bad, but we could say that if the column is named 'geography' then we assume edges are spherical? That little bit of naming of the column seems to be the only place where we can communicate a bit of metadata.

But I think we should have a clear target for this - a product that does need to export where this helps. For BigQuery we already have https://github.com/geoparquet/bigquery-converter - and the readme there makes it clear that when you try to export a geography column to Parquet it just says `Error while reading data, error message: Type GEOGRAPHY is not currently supported for parquet exports.`

I also think it's ok to have it be something for users to supply. I also think we shouldn't go too far in bending over backwards for these compatible things - we more push for tools to implement. It also seems like people could use GeoPandas pretty easily to take most any Parquet geospatial output and get it in the right shape to write out to GeoParquet.

format-specs/compatible-parquet.md
cholmes: That was @tschaub's suggestion I'm pretty sure.

format-specs/compatible-parquet.md
cholmes: So my compromise on this was to not include it in the main 'description' of a compatible file, but to just mention it as an option some readers could do. It says 'Readers can optionally support user input (in whatever manner works for the tool / library) to provide hints for metadata that is not inline with the lowest common denominator compatibility. This would include things like letting the user supply a geometry column name (ie something other than 'geometry'), using Well Known Text (WKT) in a `STRING` column instead of WKB'

I lean towards **not** including it in the core spec, since if it was there I'd want to strongly recommend people *don't* use it, since WKB is better in most ways. And it seems nicer to just give one clear target, not two that readers have to support. I see the compatible files as not 'GeoParquet', just a way to format Parquet. Like if we called it 'GeoParquet Simple' or something like that (which we decided not to for good reasons) then I could see it making a lot of sense to make sure the WKT option is in the core spec. But I'm viewing this as not so directly related, and to just suggest WKT as something readers could do to help people get data in - not even as the core 'compatibility' route.

format-specs/compatible-parquet.md
cholmes, line 3:
```suggestion
The goal of GeoParquet is that every tool producing Parquet and includes geospatial data uses official [metadata defined in the GeoParquet spec](./geoparquet_spec) to achieve true interoperability. This document represents a set of guidelines for those would like to produce geospatial Parquet data but are using tools that are not yet fully implementing GeoParquet metadata. It is meant to be used just for the interim time when only some tools properly produce GeoParquet, to enable data producers to support the growing ecosystem.
```

format-specs/compatible-parquet.md
jorisvandenbossche: OK, I am fine with keeping it mentioned here without adding it "officially" to the spec. We certainly want to avoid that libraries claim to support GeoParquet when they only or by default write WKT.

jorisvandenbossche (approved):

format-specs/compatible-parquet.md
alasarr: Yes, BigQuery throws an error when trying to export a parquet with a geography column, and as @jorisvandenbossche pointed out it's not possible to export in planar edges.

I have to say I'm not a great fan of using `geography` or `geometry` to define the edges, but I don't have a better solution, due to the importance of data warehouses in this initiative, and because Snowflake and BigQuery only support in production geography data types, I will go with the solution @cholmes proposed

The export command from BigQuery would be something like this:

```sql
EXPORT DATA
  OPTIONS (
    uri = 'gs://alasarr/geoparquet/*.parquet',
    format = 'parquet')
AS (
select ST_ASBINARY(geom) as geography FROM ...
)
```

cholmes: Cool. That reminds me one other thing that we discussed on the call that pointed towards having geometry as the default. I'm not sure if I've got it right, and @paleolimbot made the point, so he can correct me. But basically in his experience working on an s2 library it was more likely for a line defined by just two points to be a geometry - that lines with geography edges were more likely to include a number of interpolated points along the line.

alasarr: Yes, BigQuery works like that. It adds additional points to convert to geodesic: https://cloud.google.com/bigquery/docs/reference/standard-sql/geography_functions#st_geogfromgeojson

jorisvandenbossche: Do we want to add that "geometry" vs "geography" column name and default for planar vs spherical edges to the document?

alasarr: I think we should. @cholmes do you want me to create a PR for this?

cholmes: PR up at https://github.com/opengeospatial/geoparquet/pull/175

PR #175, opened 2023-07-21
updates for geography column name option

cholmes:
As discussed in https://github.com/opengeospatial/geoparquet/pull/172#discussion_r1267303288 this PR adds an option to name your column 'geography' and have readers assume your edges are spherical.


format-specs/compatible-parquet.md
```diff
@@ -26,11 +28,11 @@ The above are the key recommendations a data producer should follow. Any impleme
...
+Reading this non-compliant geospatial data from Parquet should ideally work with no user intervention if the producer followed all the guidelines and named their geometry column 'geometry'. Readers can optionally support user input (in whatever manner works for the tool / library) to provide hints for metadata that is not inline with the lowest common denominator compatibility. This would include things like letting the user supply a geometry column name (ie something other than 'geometry' or 'geography'), using Well Known Text (WKT) in a `STRING` column instead of WKB, providing a more specific geometry_type value, or providing other enhanced metadata (specifying the CRS, the winding order, the edges, the bbox or the epoch).
```
paleolimbot, line 35:
```suggestion
Reading this non-compliant geospatial data from Parquet should ideally work with no user intervention if the producer followed all the guidelines and named their geometry column 'geometry'. Readers can optionally support user input (in whatever manner works for the tool / library) to provide hints for metadata that is not inline with the lowest common denominator compatibility. This would include things like letting the user supply a geometry column name (i.e., something other than 'geometry' or 'geography'), using Well Known Text (WKT) in a `STRING` column instead of WKB, providing a more specific geometry_type value, or providing other enhanced metadata (specifying the CRS, the winding order, the edges, the bbox or the epoch).
```

paleolimbot (approved):
I haven't been following this very carefully so take my approval with a grain of salt; however, I do think this is a positive change that will result in fewer misinterpretations in the wild.

jorisvandenbossche (approved):

alasarr (approved):

Issue #178, opened 2023-07-27
Simplify or remove script dependencies


tschaub:
The release procedure currently involves running the Python scripts in the `scripts` directory to update version numbers in some example data.  The [scripts readme](https://github.com/opengeospatial/geoparquet/blob/main/scripts/README.md) suggests using Poetry to manage the scripts' dependencies.

I have just followed the instructions on installing Poetry.  After changing into the `scripts` directory and running `poetry install`, this is what I get from `poetry run python generate_example.py`:
```bash
/Users/tim/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-pt2NyYug-py3.10/lib/python3.10/site-packages/geopandas/_compat.py:111: UserWarning: The Shapely GEOS version (3.10.3-CAPI-1.16.1) is incompatible with the GEOS version PyGEOS was compiled with (3.10.1-CAPI-1.16.0). Conversions between both will be slow.
  warnings.warn(
Traceback (most recent call last):
  File "/Users/tim/projects/geoparquet/scripts/generate_example.py", line 20, in <module>
    df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
  File "/Users/tim/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-pt2NyYug-py3.10/lib/python3.10/site-packages/geopandas/io/file.py", line 166, in _read_file
    _check_fiona("'read_file' function")
  File "/Users/tim/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-pt2NyYug-py3.10/lib/python3.10/site-packages/geopandas/io/file.py", line 80, in _check_fiona
    raise ImportError(
ImportError: the 'read_file' function requires the 'fiona' package, but it is not installed or does not import correctly.
Importing fiona resulted in: dlopen(/Users/tim/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-pt2NyYug-py3.10/lib/python3.10/site-packages/fiona/ogrext.cpython-310-darwin.so, 0x0002): Library not loaded: /opt/homebrew/opt/gdal/lib/libgdal.31.dylib
  Referenced from: <8EF16756-6EC0-3D20-90D6-ECD7AE37A9E5> /Users/tim/Library/Caches/pypoetry/virtualenvs/geoparquet-scripts-pt2NyYug-py3.10/lib/python3.10/site-packages/fiona/ogrext.cpython-310-darwin.so
[...]

PR #179, opened 2023-07-27
Update scripts dependencies

kylebarron:
Closes https://github.com/opengeospatial/geoparquet/issues/178

Now 

```
poetry install
poetry run python generate_example.py
```

works on my machine


kylebarron:
This fails the tests because it uses a later version of GeoPandas/Pyproj and writes PROJJSON with a schema of `v0.6`. I created https://github.com/opengeospatial/geoparquet/issues/180

tschaub:
This should probably be a dedicated ticket, but I've run into this same problem that [the CI job is hitting](https://github.com/opengeospatial/geoparquet/actions/runs/5685503475/job/15410506979?pr=179) with the projjson schema dependency.  It is not a real problem until a version of projjson is released that is incompatible with https://proj.org/schemas/v0.5/projjson.schema.json.  But it looks like we need to relax the `test_json_schema.py` so it isn't as strict about the version that is included in the generated projjson.

kylebarron:
> But it looks like we need to relax the `test_json_schema.py` so it isn't as strict about the version that is included in the generated projjson.

It looks like pytest actually succeeds (you can run `poetry run python generate_example.py && poetry run pytest` locally; it works for me). It's these lines in the CI:

https://github.com/opengeospatial/geoparquet/blob/ad32ccaf5534a3680e62bf9de3f9837befd9c717/.github/workflows/scripts.yml#L57-L61

tschaub:
Just saw the same after re-reading.  Yes, `test -z "$(git status --porcelain)"` is the problem.

tschaub (approved):

tschaub:
Woohoo.  Thanks, @kylebarron

tschaub:
I'll merge this so we can proceed with the re-release.  Hope that sounds ok.  Thanks for fixing it up.

PR #189, opened 2023-11-17
Add GeoArrow encoding as an option to the specification

### LINKED ISSUE #185: Metadata encoding options for GeoArrow-encoded columns in GeoParquet metadata
Since #1 was opened, the https://github.com/geoarrow/geoarrow repo has seen quite a lot of exciting activity...we're getting close to releasing our initial version! We have had a lot of great anecdotal conversations about how or if geometry encoded as GeoArrow should be included in this specification and I wanted to open this issue to formalize some of the points that have been made.

Anecdotally, there has been been general agreement that including a columnar-friendly memory layout (i.e., one that does not require a parser of *any* kind to access coordinate values) as an option under the `"encoding"` metadata key would be good for GeoParquet because:

- A columnar geometry layout like GeoArrow's native encodings is closer to the in-memory form that most WKB parsers produce (i.e., it would save parsing effort)
- A columnar geometry layout may be more amenable to allowing query engines to leverage column statistics (i.e., built-in bounding boxes based on min/max field statistics) when executing spatial queries against Parquet-based file formats.
- (Let me know if I missed anything)

I think there are two orthogonal things to consider if GeoArrow will be included in a future (e.g., 1.1.0) GeoParquet specification. First, there is the question of how to structure the `"encoding"`. Currently GDAL's experimental support uses the extension type name (as summarised here: https://github.com/geoarrow/geoarrow/blob/main/extension-types.md ) as the encoding key. This is sufficient for a reader to reconstruct a GeoArrow type when reading a Parquet file:

```
{"encoding": "geoarrow.point", ...}
```

We could also just use `"geoarrow"` and declare the extension name somewhere else:

```
{"encoding": "geoarrow", "extension_name": "point"}
```

...or infer the extension name from the geometry type:

```
{"encoding": "geoarrow", ..., "geometry_types": ["Point"]}
```

(I don't like that last option because there are GeoArrow extension types for WKT and WKB. Even if they aren't necessarily allowed/encouraged for use in this spec, I don't think we can guarantee that there is one canonical extension name per combination of geometry types and functionally the extension name is what is required for a reader implementation)

The second consideration is which GeoArrow memory layouts to allow. The GeoArrow specification, like the Arrow specification, expanded to fit ways that we know people are already storing geospatial data in Arrow (i.e., it is currently more descriptive than prescriptive). The GeoParquet format and the discussions that went into creating it seem to favour a more prescriptive approach (i.e., restricting the allowed encodings/values to simplify implementations). For example, GeoParquet could provide language like:

> The only GeoArrow extensions that may be encoded in GeoParquet are geoarrow.point, geoarrow.multipoint, geoarrow.multilinestring, and geoarrow.multipolygon, and these extensions must be written using the struct/separated coordinate encoding.

The other end of the spectrum would be to just punt to the GeoArrow spec and allow any of the values we've defined.

> The extension name refers to the extension names defined in the GeoArrow specification at XXXX.

Looking forward to discussion on this! (cc @kylebarron @jorisvandenbossche )

paleolimbot:
Closes #185.

(Still in draft mode while I work on an implementation in geoarrow/geoarrow-python)


cholmes (approved):
Looks great to me!

jiayuasu (approved):
Looks good to me!

examples/example_metadata-geoarrow.json
```diff
@@ -0,0 +1,15 @@
+{
+  "geo": {
+    "columns": {
+      "geometry": {
+        "encoding": "geoarrow",
+        "extension_name": "geoarrow.point",
```
jorisvandenbossche, line 6:
```suggestion
        "geoarrow_type": "geoarrow.point",
```

format-specs/geoparquet.md
```diff
@@ -51,13 +47,14 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
...
 | orientation    | string       | Winding order of exterior ring of polygons. If present must be `"counterclockwise"`; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
 | edges          | string       | Name of the coordinate system for the edges. Must be one of `"planar"` or `"spherical"`. The default value is `"planar"`. |
 | bbox           | \[number]    | Bounding Box of the geometries in the file, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
 | epoch          | number       | Coordinate epoch in case of a dynamic CRS, expressed as a decimal year. |
+| geoarrow_type          | string       | The [GeoArrow extension name](https://geoarrow.org/extension-types#extension-names) corresponding to the column memory layout. This is required when `encoding` is `"geoarrow"` and must be omitted otherwise. |
```
jorisvandenbossche, line 57:
```suggestion
| geoarrow_type  | string       | The [GeoArrow extension name](https://geoarrow.org/extension-types#extension-names) corresponding to the column memory layout. This is required when `encoding` is `"geoarrow"` and must be omitted otherwise. |
```

jorisvandenbossche (approved):
Looks good!

It might be nice to also add a few basic tests for the json schema validation similarly as how https://github.com/opengeospatial/geoparquet/pull/191 is doing that.

format-specs/compatible-parquet.md
```diff
@@ -12,7 +12,7 @@ The core idea of the compatibility guidelines is to have the output match the de
 
 * The geometry column should be named either `"geometry"` or `"geography"`.
 
-* The geometry column should be a `BYTE_ARRAY` with Well Known Binary (WKB) used to define the geometries, as defined in the [encoding](./geoparquet.md#encoding) section of the GeoParquet spec.
+* The geometry column should be a `BYTE_ARRAY` with Well Known Binary (WKB) used to define the geometries, as defined in the [encoding](./geoparquet.md#encoding) section of the GeoParquet spec. Alternatively, the geometry column can be stored according to the Point, MultiPoint, MultiLinestring, or MultiPolygon memory layouts with separated (struct) coordinates as specified in the [GeoArrow format](https://geoarrow.org/format).
```
kylebarron, line 15:
```suggestion
* The geometry column should be a `BYTE_ARRAY` with Well Known Binary (WKB) used to define the geometries, as defined in the [encoding](./geoparquet.md#encoding) section of the GeoParquet spec. Alternatively, the geometry column can be stored according to the Point, MultiPoint, MultiLineString, or MultiPolygon memory layouts with separated (struct) coordinates as specified in the [GeoArrow format](https://geoarrow.org/format).
```
Can it not be interleaved coordinates? Is it because of XYZ/XYM confusion? I don't know how complex we want to make these suggestions, but I'd be content with a suggestion like:

> The interleaved (FixedSizeList) coordinate representation may also be used. In that case, a list size of 2 is expected to represent XY coordinates, a list size of 3 is expected to represent XYZ coordinates, and a list size of 4 is expected to represent XYZM coordinates.

examples/example_metadata-geoarrow.json
kylebarron, line 1: Nit: maybe `_geoarrow` to not mix `_` and `-`?

format-specs/schema.json
```diff
@@ -71,6 +71,10 @@
             },
             "epoch": {
               "type": "number"
+            },
+            "geoarrow_type": {
+              "type": "string",
+              "pattern": "^geoarrow\\.(point|linestring|polygon|multipoint|multilinestring|multipolygon)$"
```
kylebarron, line 77: I'm not a json schema expert, but would we be able to make this conditionally required? It looks like [`dependentRequired`](https://json-schema.org/understanding-json-schema/reference/conditionals#dependentRequired) meets what we need, though I don't know what version of json schema we're pinned to.

format-specs/geoparquet.md
kylebarron, line 57:
```suggestion
| geoarrow_type          | string       | The [GeoArrow extension name](https://geoarrow.org/extension-types#extension-names) corresponding to the column's memory layout. This is required when `encoding` is `"geoarrow"` and must be omitted otherwise. |
```

format-specs/geoparquet.md
```diff
@@ -83,10 +80,18 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
 
+Using the `"geoarrow"` encoding may provide better performance and enable readers to leverage more features of the Parquet format to accelerate geospatial queries (e.g., row group-level min/max statistics). When `encoding` is set to `"geoarrow"`, the column metadata must also specify `geoarrow_type` according to the [GeoArrow metadata specification for extension names](https://geoarrow.org/extension-types#extension-names) to signify the memory layout used by the geometry column.
+
+Note that the current version of the spec only allows for a subset of GeoArrow: separated (struct) coordinates are required, only 2D or 3D geometries are permitted, and supported extension are currently `"geoarrow.point"`, `"geoarrow.linestring"`, `"geoarrow.polygon"`, `"geoarrow.multipoint"`, `"geoarrow.multilinestring"`, and `"geoarrow.multipolygon"`. This means that M values and serialized encodings are not yet supported.
```
kylebarron, line 91: Oh I see, this doc doesn't even allow interleaved coordinates as GeoParquet. 

I'm sensitive to the complexity concerns of having too many options in the spec, but I see this as favoring the "support cloud-native remote queries" use case over the "efficient file format, but reading and writing whole tables" use case.  It "feels" like there's still a strong pull in general towards storing interleaved coordinates across the geo ecosystem.

That said, the memcopy to and from separated coordinates is pretty fast, so I can tolerate this.

format-specs/compatible-parquet.md
```diff
@@ -12,7 +12,7 @@ The core idea of the compatibility guidelines is to have the output match the de
 
 * The geometry column should be named either `"geometry"` or `"geography"`.
 
-* The geometry column should be a `BYTE_ARRAY` with Well Known Binary (WKB) used to define the geometries, as defined in the [encoding](./geoparquet.md#encoding) section of the GeoParquet spec.
+* The geometry column should be a `BYTE_ARRAY` with Well Known Binary (WKB) used to define the geometries, as defined in the [encoding](./geoparquet.md#encoding) section of the GeoParquet spec. Alternatively, the geometry column can be stored according to the Point, MultiPoint, MultiLineString, or MultiPolygon memory layouts with separated (struct) coordinates as specified in the [GeoArrow format](https://geoarrow.org/format).
```
Kontinuation, line 15: Does this mean that geometry column with mixed types of geometries cannot be encoded as GeoArrow?

kylebarron: Not until https://github.com/apache/parquet-format/pull/44 is merged (if ever)

paleolimbot: It can (or will be once we sort out the details of https://github.com/geoarrow/geoarrow/pull/43 ), although it's unclear exactly how we'd do that in Parquet or if it would be useful in Parquet. In any case, it would be a future addition!

Kontinuation: Columns with mixed geometry values are quite common for most query engines with geospatial support. Most of the time geometry columns have the umbrella type "geometry" or "geography", and it is not practical to first resolve the subtypes of the geometries before writing out parquet files. I'd look forward to a columnar encoding supporting mixed geometry types as well as geometry collections.

kylebarron: Arrow and Parquet are two different specs. Arrow has a union type, which allows for mixed geometry types in a single column, while maintaining constant-time access to any coordinate. Parquet does not today have a union type, so it's impossible to to write the `Geometry` and `GeometryCollection` arrays in https://github.com/geoarrow/geoarrow/pull/43 natively to Parquet.

GeoArrow implementations are able to statically know whether they have singly-typed geometries in a column, in which case they can write one of the 6 primitive types. GeoArrow implementations will have to fall back to WKB-encoded geometries for mixed-type columns. I don't see how this is something we could realistically change, unless we essentially [re-implement union handling in a struct](https://github.com/geoarrow/geoarrow/pull/43#pullrequestreview-1560131039), which would be a big ask for implementors.

Kontinuation: I understand that Arrow and Parquet are different specs, and it is impossible to write mixed-typed geometry values to Parquet files using GeoArrow encoding. So do we have to define yet another encoding and add it to the spec if we want to support encoding mixed-typed geometry values in Parquet in a columnar fashion (preserving all the benefits of row-group and page level statistics)? The new encoding could be something similar to `Struct<geometry_type: int8, multipolygon>` or [Spatial Parquet](https://arxiv.org/abs/2209.02158).

kylebarron: > So do we have to define yet another encoding and add it to the spec if we want to support encoding mixed-typed geometry values in Parquet in a columnar fashion (preserving all the benefits of row-group and page level statistics)? 

No, we can have a WKB column plus a bounding box column (https://github.com/opengeospatial/geoparquet/pull/191).

> The new encoding could be something similar to `Struct<geometry_type: int8, multipolygon>`

This is what @paleolimbot offered in https://github.com/geoarrow/geoarrow/pull/43#pullrequestreview-1560131039 and what I referred to as "re-implement union handling in a struct". 

In any case, I'd argue that a large majority of use cases will be solved by these 6 primitive types, and we can come back to union types in the future.

jiayuasu: @paleolimbot and @kylebarron Thanks for leading the effort on this. This is great work.

I believe @Kontinuation 's point is more like: without allowing mixed geometry types in a column, all database engines that didn't use Arrow as the in-memory format have no way to know if a geometry column is a mixed type or not. In other words, they don't know if they can use this GeoArrow encoding to GeoParquet files.

The only way to know it is that scanning every single record of a big dataset first (get all geometry types), then in the second round, write the table to GeoParquet. This process will significant slow down the write performance and add challenges to the implementer or database users.

***So, regardless if the 6 primitive types cover the majority of the use cases or not, as long as a database engine wants to write GeoArrow encoding in GeoParquet, this additional check is inevitable.***

* Many database engines are gonna be affected by this: Snowflake, Big Query, AWS Redshift, Apache Sedona, Trino. Because their Geometry/Geography column allows mixed geometries.

* Engines will not be affected by this: DataFusion, GeoPandas (with Arrow), or any engine that uses Arrow/GeoArrow as its in-memory layout.

On the other hand, according to https://github.com/apache/parquet-format/pull/44, the Union type of Parquet might never be accepted by the Parquet community given that Apache Iceberg also does not allow Union type. Multiple Apache Iceberg PMCs are Apache Parquet PMCs as well. ***This means it is very likely that we will have to `re-implement union handling in a struct`***

That said, I think we can still add the GeoArrow encoding to the GeoParquet spec since I don't think this will create much incompatibility if we fix this issue in the future.

***The only downside of adding this to the spec now is that the adoption of writing GeoArrow encoding in GeoParquet might be low.***

Just my 2 cents. Please feel free to correct me if I am wrong.

paleolimbot: All good points!

I think Kyle put the summary best:

> In any case, I'd argue that a large majority of use cases will be solved by these 6 primitive types, and we can come back to union types in the future.

In the context of this PR, that would mean that the column option `geoarrow_type` could in the future be set to `"geoarrow.mixed"`.

I don't think we anticipated that writing mixed geometries in geoarrow to Parquet would be the main use-case. If this *is* an important use, please chime in on https://github.com/geoarrow/geoarrow/pull/43 with some details! We definitely don't want to represent mixed geometries in a way that prevents them being used.

> The only way to know it is that scanning every single record of a big dataset first (get all geometry types), then in the second round, write the table to GeoParquet

This is only true if there's no ability for a user to supply any information about the encoding. If there is, `write_geoparquet(..., encoding = geoarrow("geoarrow.mixed"))` should do it. Typically the user does know this information (even if the database does not).

kylebarron: This discussion is making me think that the GeoParquet spec should not be defined _in terms of_ GeoArrow. Rather it should be defined as "native type" or "flat type" or similar. Then a sentence in prose can mention that it overlaps partially with the GeoArrow spec.

I'm also becoming convinced that the serialized form need not exactly overlap with GeoArrow. On the topic of mixed arrays specifically, as possibly the only one who has written an implementation of GeoArrow mixed arrays, I've become an even stronger proponent of using an Arrow union type for GeoArrow mixed arrays because of its ability to know geometry types statically. So I think the best way forward for GeoParquet (for a future PR) would be to discuss a "struct-union" approach for GeoParquet that is _not_ the same in-memory representation as GeoArrow.

> all database engines that didn't use Arrow as the in-memory format have no way to know if a geometry column is a mixed type or not. In other words, they don't know if they can use this GeoArrow encoding to GeoParquet files

I think changing nomenclature will also be clearer to non-arrow-based implementations that reading and writing this "native" encoding of GeoParquet is _not_ dependent on using Arrow internally.

So my recommendation would be to **take out most references to geoarrow** from this PR. I.e., we don't want the metadata key called `geoarrow_type` if there's a possibility where the GeoParquet union type is not the same as the GeoArrow union type.

paleolimbot: I actually think the strength of this PR is the strong delegation to the GeoArrow specification: I don't think we should be developing two specifications, particularly since we have very little engagement on the GeoArrow spec already. We've spent quite a bit of time documenting the memory layouts for geoarrow in that specification and I don't think it would be productive to copy/paste those here and maintain them independently. I also don't think it would be productive to link to the GeoArrow specification for documentation of all the memory layouts but very pointedly *not* call it GeoArrow.

It may be that representing mixed geometry is not important in the context of GeoParquet (Maybe WKB is just as fast in the context of compression + IO + Parquet's list type? Have we checked?), or it may be that there is a common memory representation that makes sense for both specifications that will improve interoperability (although that would involve quite a lot of reimplementation on Kyle's end 😬 ).

I don't want us to loose track of the main point here, which is that this PR is mostly about enabling very efficient representations of single-type geometries, which are *very* commonly the types of files that you might want to put in a giant S3 bucket and scan efficiently.

jorisvandenbossche: > This discussion is making me think that the GeoParquet spec should not be defined _in terms of_ GeoArrow. Rather it should be defined as "native type" or "flat type" or similar. Then a sentence in prose can mention that it overlaps partially with the GeoArrow spec.

We could go back to to the question about which "encoding" values to allow, and instead of a generic `"geoarrow"` option (with an additional "geoarrow_type" key to then be more specific), have actual encoding options `"point"`, `"linestring"`, `"polygon"`, etc.

For the rest, it is mostly a question about how to document this: how to phrase this exactly in the specification, how strongly to tie it to geoarrow (or just reference as mostly similar), how much to duplicate the details of the memory layout, etc. But that's all "details" about the best way to document it, while keeping the actual specification (what ends up in the metadata in a file) agnostic to geoarrow. 

I think I am somewhat convinced by @kylebarron's points on this, and like to idea of having the actual spec changes not use "geoarrow" (and then we can still debate how much to use the term in the explanation of it).

For example, as long as the specification would exactly overlap (or be a strict subset of geoarrow), we can still point to the geoarrow spec for the details to avoid too much duplication (and having to maintain two versions). And this is also easy to change then in the future if we would want to introduce differences.

>  I also don't think it would be productive to link to the GeoArrow specification for documentation of all the memory layouts but very pointedly not call it GeoArrow.

On the other hand, for an implementation of GeoParquet in some library that has nothing to do with Arrow (doesn't use an Arrow implementation under the hood), the "geoarrow" name is also somewhat uninformative, when strictly looking at it from a GeoParquet point of view.

paleolimbot: Perhaps it's worth crafting a new PR that uses the language you all are hoping for with a draft implementation? I don't currently have the bandwidth to do that but am happy to review!

format-specs/geoparquet.md
```diff
@@ -12,11 +12,7 @@ This is version 1.1.0-dev of the GeoParquet specification.  See the [JSON Schema
 
 ## Geometry columns
 
-Geometry columns MUST be stored using the `BYTE_ARRAY` parquet type. They MUST be encoded as [WKB](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary).
```
jorisvandenbossche, line 15: Can we keep the information of the first sentence somewhere? (i.e. that for a WKB encoding, the geometry column MUST be stores using the `BYTE_ARRAY` parquet type)

(you kept the "Implementation note" just below that also mentions `BYTE_ARRAY`, but that is not so specific as the above)

examples/example_metadata_geoarrow.json
```diff
@@ -0,0 +1,15 @@
+{
+  "geo": {
+    "columns": {
+      "geometry": {
+        "encoding": "geoarrow",
+        "geoarrow_type": "geoarrow.point",
+        "geometry_types": [
+          "Point"
+        ]
```
kylebarron, line 9: This still needs to be updated.

kylebarron (approved):

format-specs/geoparquet.md
paleolimbot: Done! We could also add some more in here about the Parquet physical description of how nesting works (but maybe in a future PR?)

examples/example_metadata_geoarrow.json
paleolimbot: 😬  (Done!)

format-specs/geoparquet.md
```diff
@@ -82,15 +82,25 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
+WKB geometry columns MUST be stored using the `BYTE_ARRAY` parquet type.
 
 Implementation note: when using WKB encoding with the ecosystem of Arrow libraries, Parquet types such as `BYTE_ARRAY` might not be directly accessible. Instead, the corresponding Arrow data type can be `Arrow::Type::BINARY` (for arrays that whose elements can be indexed through a 32-bit index) or `Arrow::Type::LARGE_BINARY` (64-bit index). It is recommended that GeoParquet readers are compatible with both data types, and writers preferably use `Arrow::Type::BINARY` (thus limiting to row groups with content smaller than 2 GB) for larger compatibility.
 
+##### Geometry type specific encodings (based on GeoArrow)
```
kylebarron, line 100: Maybe
```
##### Native encodings (based on GeoArrow)
```
?

jorisvandenbossche: Yeah, I don't know what's the best terminology to use here. "Geometry type specific" might also not be true in the future if we add a union-like version.

I was a bit hesitant to use "native" because that's also vague / can be interpreted in multiple ways, but happy to use this if others think that is better

kylebarron: I like the distinction between "serialized" and "native" encodings. "Geometry type specific encodings" is a mouthful for a header

format-specs/geoparquet.md
```diff
@@ -83,10 +81,68 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
+For the other geometry types, those x and y coordinate values MUST be embedded
+in repeated groups (`LIST` logical parquet type). For example, for the
+`"multipolygon"` geometry type:
+
+```
+// "multipolygon" geometry column with multiple levels of nesting
+optional group geometry (List) {
+  // the parts of the MultiPolygon
+  repeated group list {
+    optional group element (List) {
```
jorisvandenbossche, line 128: This is verbose, but this matches how the Parquet docs specify a List type: https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists

In the Spatial Parquet paper, they made this shorter (https://arxiv.org/pdf/2209.02158.pdf), and essentially something like (translating their syntax to our layout):

```
optional group geometry{
  repeated group part {
    repeated group coordinate {
      required double x;
      required double y;
    }
  }
}
```

While that is clearer / more readable for the reader not super familiar with Parquet details, it also has value to stick with the way the Parquet docs do this (and eg how printing the Parquet schema of a file with pyarrow will also do this).

format-specs/geoparquet.md
```diff
@@ -83,10 +81,68 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
+      // the rings of one Polygon
+      repeated group list {
+        optional group element (List) {
```
jorisvandenbossche, line 131: Another element here is that the Parquet spec says to use the generic "list" and "element" field names, which means we don't use the more descriptive "polygons" / "rings" / "vertices" names as in GeoArrow

(until recently, pyarrow would honor the names in the Arrow schema and write a "non-compliant" Parquet file, but now it defaults to writing a compliant one, which you can still turn off with `use_compliant_nested_type=False`)

format-specs/geoparquet.md
```diff
@@ -83,10 +81,68 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
+          // the list of coordinates of one ring
+          repeated group list {
+            optional group element {
+              required double x;
+              required double y;
```
jorisvandenbossche, line 136: At the moment I only made the inner x and y as "required" here, but I _think_ I should also mark the other sub-levels (for polygons/rings/vertices) as required? I think we said only the geometry itself can be null in the GeoArrow spec.

jorisvandenbossche: Changed all to "required" except for the top-level geometry group

paleolimbot: I think it would be a little strange if the schema written by pyarrow for a GeoArrow encoded array didn't match the text here. Does marking them as required have a storage optimization?

paleolimbot: We can probably make sure that everything except the outer layer is written as non-nullable fields regardless of how GeoArrow deals with that. In the proof-of-concept there's a step where the extension information is stripped, at which point we could also set the non-nullable-ness if this really does matter.

Once wrapped in an extension type, I don't think the explicit non-nullable-ness of the storage is all that useful. The assumptions of buffer data are governed by the extension type, and we say there are no nulls allowed except at the outer level.

kylebarron: How much is the data saving? Does parquet essentially not need a nullability buffer for these arrays when the child arrays are saved as non-nullable? In my own implementation I think I'm pretty strict about having child fields marked as non-nullable, so I'm ok making this required here if you are.

paleolimbot: I forget what the barrier was when we last revisited this...I think it was maybe ensuring storage type equality or the ability of various implementations to actually create a list with a non-nullable child. In any case, we should make the spec clear what is allowed in terms of a storage type. I'm not convinced that there's a practical application to non-nullable children but I'm game to update geoarrow-c to reflect whatever is in the spec (and happy to export whatever Parquet schema is decided on here).

jorisvandenbossche: > How much is the data saving? Does parquet essentially not need a nullability buffer for these arrays when the child arrays are saved as non-nullable?

Looking at the Parquet docs, it says (https://parquet.apache.org/docs/file-format/nulls/):

> Nullity is encoded in the definition levels (which is run-length encoded). 

and then here (https://parquet.apache.org/docs/file-format/data-pages/):

> For data that is required, the definition levels are skipped

So essentially if a field is required, it can skip some data being stored (the definition levels for that field). Although given that it is run-length encoded, this data will also be tiny in case there are no nulls (because then essentially it's just one constant being run-length encoded, I assume)

See also https://arrow.apache.org/blog/2022/10/05/arrow-parquet-encoding-part-1/

jorisvandenbossche: For the actual spec here, I think we can recommend `required` (and show it like that in the example) but mention that it is not required and that using `optional` fields if also fine (i.e. it's not a MUST, for example in case there is no easy way to control it depending on the Parquet implementation you're using)

kylebarron: > For the actual spec here, I think we can recommend `required` (and show it like that in the example) but mention that it is not required and that using `optional` fields if also fine (i.e. it's not a MUST, for example in case there is no easy way to control it depending on the Parquet implementation you're using)

Yeah that sounds fine to me. Maybe say something that there MUST be no null values for child arrays, even if it's marked as `optional`?

format-specs/geoparquet.md
```diff
@@ -83,10 +81,75 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
+            }
+          }
+        }
+      }
+    }
+  }
+}
+```
+
+There MUST NOT be any null values in the child fields and the x/y coordinate
```
paleolimbot, line 146:
```suggestion
There MUST NOT be any null values in the child fields and the x/y/z coordinate
```

(Right? We support Z here?)

format-specs/geoparquet.md
```diff
@@ -83,10 +81,75 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
+fields. Only the outer optional "geometry" group is allowed to have nulls (i.e
+representing a missing geometry). This MAY be indicated in the Parquet schema by
+using `required` group elements, as in the example above, but this is not
+required and `optional` fields are fine as well (as long as the data itself does
```
paleolimbot, line 150:
```suggestion
required and `optional` fields are permitted (as long as the data itself does
```

paleolimbot (commented):
Thank you for adding the Parquet type information!

cholmes (approved):

rouault:
Are there (small) Parquet datasets somewhere using this new GeoArrow encoding? I now realize that the current support for GeoArrow in GDAL used the FixedSizeList[2] encoding and not the struct one... So I need to do changes and check interoperability

format-specs/geoparquet.md
```diff
@@ -83,10 +81,75 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
+
+For the other geometry types, those x and y coordinate values MUST be embedded
+in repeated groups (`LIST` logical parquet type). For example, for the
+`"multipolygon"` geometry type:
+
+```
+// "multipolygon" geometry column with multiple levels of nesting
+optional group geometry (List) {
+  // the parts of the MultiPolygon
+  repeated group list {
```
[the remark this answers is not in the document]
jorisvandenbossche: I mentioned this issue above at https://github.com/opengeospatial/geoparquet/pull/189#discussion_r1520080717

The Parquet spec specifies that those names should be "list" and "element": https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists 

But it also mentions that for backwards compatibility you should be able to read lists with other names:

> It is required that the repeated group of elements is named `list` and that its element field is named `element`. However, these names may not be used in existing data and should not be enforced as errors when reading.

Until recently, PyArrow / Arrow C++ actually created such "non-compliant" files, but there is a flag to control this, for which the default has been changed to create compliant files by default since pyarrow 13.0 (https://github.com/apache/arrow/issues/29781).

Illustrating this:

```python
point_type = pa.struct([pa.field("x", pa.float64(), nullable=False), pa.field("y", pa.float64(), nullable=False)])
polygon_type = pa.list_(
    pa.field("rings", pa.list_(pa.field("vertices", point_type, nullable=False)), nullable=False)
)
multipolygon_type = pa.list_(pa.field("polygons", _polygon_type, nullable=False))

table = pa.table({"geometry": pa.array([[[[{'x': 1.0, 'y': 2.0}, {'x': 2.0, 'y': 3.0}]]]], type=multipolygon_type)})

>>> pq.write_table(table, "test_multipolygon_non_nullable-compliant.parquet")
>>> pq.read_metadata("test_multipolygon_non_nullable-compliant.parquet").schema
<pyarrow._parquet.ParquetSchema object at 0x7efda2333b40>
required group field_id=-1 schema {
  optional group field_id=-1 geometry (List) {
    repeated group field_id=-1 list {
      required group field_id=-1 element (List) {
        repeated group field_id=-1 list {
          required group field_id=-1 element (List) {
            repeated group field_id=-1 list {
              required group field_id=-1 element {
                required double field_id=-1 x;
                required double field_id=-1 y;
              }
            }
          }
        }
      }
    }
  }
}

>>> pq.write_table(table, "test_multipolygon_non_nullable-non_compliant.parquet", use_compliant_nested_type=False)
>>> pq.read_metadata("test_multipolygon_non_nullable-non_compliant.parquet").schema
<pyarrow._parquet.ParquetSchema object at 0x7efda135fd00>
required group field_id=-1 schema {
  optional group field_id=-1 geometry (List) {
    repeated group field_id=-1 list {
      required group field_id=-1 polygons (List) {
        repeated group field_id=-1 list {
          required group field_id=-1 rings (List) {
            repeated group field_id=-1 list {
              required group field_id=-1 vertices {
                required double field_id=-1 x;
                required double field_id=-1 y;
              }
            }
          }
        }
      }
    }
  }
}
```

Specifically for Arrow C++ (or other implementation that do this), we do store the original Arrow schema in the Parquet metadata, and so we could preserve the custom names on roundtrip. But it seems we are not doing that currently.

But this actually raises the question: for the GeoArrow spec itself, how "required" are those custom field names? Are you supposed to rename them after reading a compliant Parquet file that will not have those names? 
At the moment the GeoArrow spec itself seems to mention "should" for the names (https://github.com/geoarrow/geoarrow/blob/main/format.md#memory-layouts)

> Thinking that for bounding box stats I'll need to find the leaf arrays by name and for multipolygon it'll be something like `geometry.list.list.list.x` and ``geometry.list.list.list.y`?

To get back to that initial question: it will be something like `geometry.list.element.list.element.x` (with the number of `list.element` repetitions depending on the geometry type)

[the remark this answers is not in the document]
kylebarron: > But it also mentions that for backwards compatibility you should be able to read lists with other names:

👍 

> But this actually raises the question: for the GeoArrow spec itself, how "required" are those custom field names? Are you supposed to rename them after reading a compliant Parquet file that will not have those names? 

My implementation will likely rename field names to the GeoArrow prescribed ones.

> To get back to that initial question: it will be something like `geometry.list.element.list.element.x` (with the number of `list.element` repetitions depending on the geometry type)

👍

rouault:
@paleolimbot I believe there's a slight non-conformity with the schema of your samples attached to https://github.com/opengeospatial/geoparquet/pull/189#issuecomment-2020497791. The x/y fields of the struct are currently marked as optional, whereas the spec requires them to be required. Cf
```
$ ~/arrow/cpp/build/release/parquet-dump-schema examples/point-geoarrow.parquet
required group field_id=-1 schema {
  optional int32 field_id=-1 row_number;
  optional group field_id=-1 geom {
    optional double field_id=-1 x;
    optional double field_id=-1 y;
  }
}
```

jorisvandenbossche:
While we should provide some example data that fully follows the recommendations, those files are compliant in the sense that we recommend marking those fields as `required`, but that's not required (`optional` is fine as well):

> There MUST NOT be any null values in the child fields and the x/y/z coordinate
fields. Only the outer optional "geometry" group is allowed to have nulls (i.e
representing a missing geometry). This MAY be indicated in the Parquet schema by
using `required` group elements, as in the example above, but this is not
required and `optional` fields are permitted (as long as the data itself does
not contain any nulls).

rouault:
> those files are compliant in the sense that we recommend marking those fields as `required`, but that's not required (`optional` is fine as well):

ah ok, missed that. Sorry for the noise.

jorisvandenbossche:
We should probably also provide a version of the included `example.parquet` in geoarrow encoding, as well as for the `[nz-building-outlines.parquet](https://storage.googleapis.com/open-geodata/linz-examples/nz-building-outlines.parquet)`.

For geoarrow-encoding, given that this is geometry type specific, it might make sense to include example files in this repo for all types. Or update the [geoarrow-data](https://github.com/geoarrow/geoarrow-data) repo to also include Parquet versions of the files (currently I think it hosts only .arrow files?)

PR #190, opened 2023-11-17
Spec: add a note about mapping of Parquet BYTE_ARRAY type to Arrow types

### LINKED ISSUE #187: Recommendation on the Arrow specific type for the WKB geometry column ?
The GeoParquet spec rightly specfies the type of geometry columns in terms of Parquet type: "Geometry columns MUST be stored using the BYTE_ARRAY parquet type"
For implementations using the Arrow library (typically from C++, but possibly from Python or other languages), they might use the Arrow API for Parquet reading & writing and thus not be directly exposed to Parquet types.
I came to realize recently that the GDAL implementation would only work for the Arrow::Type::BINARY, but not for Arrow::Type::LARGE_BINARY. This has been addressed, on the reading side of the driver, for GDAL 3.8.0 per https://github.com/OSGeo/gdal/pull/8618 .
I'm not entirely clear how the Arrow library maps a large Parquet file with row group with a WKB column with more than 2 GB of content. I assume that would be Arrow::Type::LARGE_BINARY ?
So the question is if there should be some hints in the spec for implementations using the Arrow library on:
- which Arrow types they should expect on the reading side
- which Arrow type(s) they should use on the writing side

rouault:
Fixes #187

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -13,6 +13,9 @@ This is version 1.1.0-dev of the GeoParquet specification.  See the [JSON Schema
 ## Geometry columns
 
 Geometry columns MUST be stored using the `BYTE_ARRAY` parquet type. They MUST be encoded as [WKB](https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Well-known_binary).
+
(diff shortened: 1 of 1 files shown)


paleolimbot (approved):
This wording seems reasonable to me! Thank you!

kylebarron (approved):

jorisvandenbossche:
Thanks Even! That looks good to me as well. We should probably move that to the encoding section when there will be multiple encodings, but that can be done in a later PR.

PR #196, opened 2024-01-30
Fixes the link from `compatible-parquet.md` to the schema.

xaviernogueira:
The link was broken before, pointing to a non existent "geoparquet_spec" file. I changed it to point to the `schema.json`.


tschaub (approved):
Thanks, @xaviernogueira

PR #202, opened 2024-04-15
Bounding box column updates

kylebarron:
- More clearly specify that the bounding box column may have 4 or 6 fields: e.g. that 3D bounding boxes are allowed
- Strictly specify the ordering of the fields within the struct. 
- More clearly specify that if `zmin` is included, then `zmax` must also be included, and vice versa.


jorisvandenbossche (approved):

format-specs/geoparquet.md
```diff
@@ -227,7 +227,7 @@ Note: This technique to use the bounding box to improve spatial queries does not
...
+A bounding box column MUST be a Parquet group field with 4 or 6 child fields representing the geometry's coordinate range. For two-dimensional data, the child fields MUST be named `xmin`, `ymin`, `xmax`, and `ymax` and MUST be ordered in this same way. As with the top-level [`bbox`](#bbox) column, the values follow the GeoJSON specification (RFC 7946, section 5), which also describes how to represent the bbox for geometries that cross the antimeridian. For three dimensions the additional fields `zmin` and `zmax` MAY be present but are not required. If `zmin` is present then `zmax` MUST be present and vice versa. If `zmin` and `zmax` are present, the ordering of the child fields MUST be `xmin`, `ymin`, `zmin`, `xmax`, `ymax`, `zmax`. The fields MUST be of Parquet type `FLOAT` or `DOUBLE` and all columns MUST use the same type. The repetition of a bounding box column MUST match the geometry column's [repetition](#repetition). A row MUST contain a bounding box value if and only if the row contains a geometry value. In cases where the geometry is optional and a row does not contain a geometry value, the row MUST NOT contain a bounding box value.
```
tschaub, line 230: I notice that this requirement about the ordering of the fields makes invalid the newly released `2024-04-16-beta.0` [Overture Maps data](https://github.com/OvertureMaps/data) (where the fields in the `bbox` group happen to be ordered `xmin`, `xmax`, `ymin`, and `ymax`).

I've missed recent meetings, and maybe this has been covered, but I'm curious about including both the name and the order of the fields in the spec - this feels "overspecified" to me.  Are there clients that cannot work with the named fields in a group?

I don't know of one, but it wouldn't surprise me if there were data producers that couldn't control the order of the fields in a group (maybe using a language where unordered struct fields are serialized in a Parquet group field).

m-mohr: I'm also wondering why the ordering is required? Shouldn't access go through the names that are specified in the array?

kylebarron: I think the difference here is whether you consider the struct type as a hashmap or as an ordered list of tuples. I see Parquet's struct type more as the latter than the former. E.g. the struct type is really the collection of fields like so

```
[
	("bbox", "xmin"),
	("bbox", "ymin"),
	("bbox", "xmax"),
	("bbox", "ymax")
]
```

You can loop through the columns to find the column index for your desired name, but it's not a map. You can't directly access a column by name. Instead, you must first find the column index, then access the column by index.

Given that low-level access is done by index, why not strictly specify column order in the spec as well? If there were some writer that was unable to specify column order in a struct type, that might be a good reason. Does that exist?

cholmes:
@maxxen - is there any problems for DuckDB if we did this proposed strict specification of order in the struct?

Maxxen:
@cholmes I can't think of any reason why the ordering would matter for us right now, so feel free to go ahead! 

( I also kind of like: xmin, xmax, ymin, ymax [zmin, zmax] as the optional axis won't be interleaved)

kylebarron:
> ( I also kind of like: xmin, xmax, ymin, ymax [zmin, zmax] as the optional axis won't be interleaved)

We had a discussion today about this on the call and various people supported that.

jorisvandenbossche:
The main annoyance of using `xmin, xmax, ymin, ymax [, zmin, zmax]` is that this is then inconsistent with the `bbox` column-level metadata, where we require an array of `[<xmin>, <ymin>, <xmax>, <ymax>]` or `[<xmin>, <ymin>, <zmin>, <xmax>, <ymax>, <zmax>]` ... (https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#bbox)

It's of course a different kind of data structure, the latter is a json array, and what we are discussing here is just the order of fields in the schema. But it would still have been nice to use the same order. Although I definitely agree that `xmin, xmax, ymin, ymax [, zmin, zmax]` sounds easier to handle

rouault:
> But it would still have been nice to use the same order.

+1

Maxxen:
Alright, well if thats already standardized then maybe it makes more sense to follow the existing convention.

cholmes (approved):

Issue #199, opened 2024-03-26
Update example files for 1.1


cholmes:
We should probably also provide a version of the included `example.parquet` in geoarrow encoding, as well as for the [nz-building-outlines.parquet](https://storage.googleapis.com/open-geodata/linz-examples/nz-building-outlines.parquet).

For geoarrow-encoding, given that this is geometry type specific, it might make sense to include example files in this repo for all types. Or update the [geoarrow-data](https://github.com/geoarrow/geoarrow-data) repo to also include Parquet versions of the files (currently I think it hosts only .arrow files?)

_Originally posted by @jorisvandenbossche in https://github.com/opengeospatial/geoparquet/issues/189#issuecomment-2020578039_

kylebarron:
I think it makes sense to include some (very small) data files in this repo, or in a specific geoparquet-data repo

paleolimbot:
I agree that the canonical place for that test data should be something geoparquet-specific (although I also think it makes sense for geoarrow-data to distribute some files too since we're already distributing multiple formats).

cholmes:
So we've got #204 in - is there more we want to do here? Get #134 in?

jorisvandenbossche:
I think it would also be good to have the example.parquet file in a different encoding

PR #204, opened 2024-04-16
Add test data covering different native (geoarrow-based) encodings

jorisvandenbossche:
Addressing part of https://github.com/opengeospatial/geoparquet/issues/199 (small example data)

And similar as https://github.com/opengeospatial/geoparquet/pull/134, but focusing here on the new geoarrow-based encodings (I should also update that older PR)

This adds some tiny example data for the different geometry types:

- for each I am saving the data as WKB and the native encoding (with some default metadata), and additionally also as plain text in case this would be useful for another tool as the source of truth to compare with (eg to check reading the parquet file was correct)
- this scripts only relies on pyarrow fro creating the data and writing the parquet files (and shapely to create the WKB data), so nothing geoarrow-specific (eg no arrow extension types, no interleaved coordinates, so just what is in the geoparquet spec)

In https://github.com/opengeospatial/geoparquet/pull/134, I also saved the metadata in a separate file as well, which allows easier validation / review that this is correct. I could do that here as well, in case that is useful.
(I should still add validation of the generated files with the json schema)

Does this look useful in this form? Any suggestions to approach it differently?


test_data/data-linestring.wkt
paleolimbot, line 1: These look like .csv files to me...should they have a .csv extension?

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+    table = table.replace_schema_metadata({"geo": json.dumps(metadata)})
+    pq.write_table(table, HERE / f"data-{geometry_type.lower()}-encoding_native.parquet")
+
+
+# point
+
+geometries_wkt = [
+    "POINT (30 10)",
+    "POINT EMPTY",
+]
```
paleolimbot, line 67: Should there also be a version of these that contain NULLs or a version that contains Z values?

paleolimbot (commented):
Just a few notes! (I haven't checked the actual files yet)

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+# linestring
+
+geometries_wkt = [
+    "LINESTRING (30 10, 10 30, 40 40)",
+    "LINESTRING EMPTY",
+]
+
+linestring_type = pa.list_(pa.field("vertices", point_type, nullable=False))
+geometries = pa.array(
+    [[(30, 10), (10, 30), (40, 40)], []], type=linestring_type)
```
kylebarron, line 90: Maybe it would be less error-prone to construct these geometries from the WKT? Then we don't have to manually check that the WKT is equivalent to here. 

_Or_ maybe it would be better to use GeoJSON and pyarrow's JSON reader to infer geometry columns? Though then you'd need to convert from interleaved to separated coords, which is a bit involved for this test data here 😕

kylebarron (approved):

test_data/generate_test_data.py
jorisvandenbossche: Yeah, I actually tried to automate writing this hard-coded data a bit, by parsing the WKT with shapely and converting to GeoJSON. The main problem is that then I need to convert the inner lists to tuples (as json is only lists), to have the construction as a separated struct type work with pyarrow (a tuple element can be converted to a struct, but a list element not). And that gives some convoluted logic ..

jorisvandenbossche:
cc @rouault would this (have been) useful for testing in GDAL? (I have been checking the files generated here with GDAL main, and they read fine)

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+HERE = pathlib.Path(__file__).parent
+
+
+metadata_template = {
+    "version": "1.1.0",
+    "primary_column": "geometry",
+    "columns": {
+        "geometry": {
+            "encoding": "WKB",
+            "geometry_type": [],
```
rouault, line 31:
```suggestion
            "geometry_types": [],
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+        },
+    },
+}
+
+
+## Various geometry types with WKB and native (GeoArrow-based) encodings
+
+def write_encoding_files(geometries_wkt, geometries_geoarrow, geometry_type):
```
rouault, line 39:
```suggestion
def write_encoding_files(geometries_wkt, geometries_geoarrow, geometry_types):
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+
+    table = pa.table({"col": range(len(geometries_wkt)), "geometry": geometries_wkt})
+    write_csv(table, HERE / f"data-{geometry_type.lower()}.wkt")
+
+    # WKB encoding
+    table = pa.table(
+        {"col": range(len(geometries_wkt)), "geometry": to_wkb(from_wkt(geometries_wkt))}
+    )
+    metadata = copy.deepcopy(metadata_template)
+    metadata["columns"]["geometry"]["geometry_type"] = [geometry_type]
```
rouault, line 49:
```suggestion
    metadata["columns"]["geometry"]["geometry_types"] = [geometry_types]
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+    table = table.replace_schema_metadata({"geo": json.dumps(metadata)})
+    pq.write_table(table, HERE / f"data-{geometry_type.lower()}-encoding_wkb.parquet")
```
rouault, line 51:
```suggestion
    pq.write_table(table, HERE / f"data-{geometry_types.lower()}-encoding_wkb.parquet")
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+
+    # native (geoarrow) encoding
+    table = pa.table(
+        {"col": range(len(geometries_wkt)), "geometry": geometries_geoarrow}
+    )
+    metadata["columns"]["geometry"]["encoding"] = geometry_type.lower()
```
rouault, line 57:
```suggestion
    metadata["columns"]["geometry"]["encoding"] = geometry_types.lower()
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+    table = table.replace_schema_metadata({"geo": json.dumps(metadata)})
+    pq.write_table(table, HERE / f"data-{geometry_type.lower()}-encoding_native.parquet")
```
rouault, line 59:
```suggestion
    pq.write_table(table, HERE / f"data-{geometry_types.lower()}-encoding_native.parquet")
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+    "LINESTRING (30 10, 10 30, 40 40)",
+    "LINESTRING EMPTY",
+]
+
+linestring_type = pa.list_(pa.field("vertices", point_type, nullable=False))
+geometries = pa.array(
+    [[(30, 10), (10, 30), (40, 40)], []], type=linestring_type)
+
+write_encoding_files(
+    geometries_wkt, geometries, geometry_type="LineString"
```
rouault, line 93:
```suggestion
    geometries_wkt, geometries, geometry_types="LineString"
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+point_type = pa.struct(
+    [
+        pa.field("x", pa.float64(), nullable=False),
+        pa.field("y", pa.float64(), nullable=False)
+    ]
+)
+geometries = pa.array([(30, 10), (float("nan"), float("nan"))], type=point_type)
+
+write_encoding_files(
+    geometries_wkt, geometries, geometry_type="Point"
```
rouault, line 78:
```suggestion
    geometries_wkt, geometries, geometry_types="Point"
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+        [[(30.0, 10.0), (40.0, 40.0), (20.0, 40.0), (10.0, 20.0), (30.0, 10.0)]],
+        [[(35.0, 10.0), (45.0, 45.0), (15.0, 40.0), (10.0, 20.0), (35.0, 10.0)],
+         [(20.0, 30.0), (35.0, 35.0), (30.0, 20.0), (20.0, 30.0)]],
+        [],
+    ],
+    type=polygon_type
+)
+
+write_encoding_files(
+    geometries_wkt, geometries, geometry_type="Polygon"
```
rouault, line 120:
```suggestion
    geometries_wkt, geometries, geometry_types="Polygon"
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+    [
+        [(30.0, 10.0)],
+        [(10.0, 40.0), (40.0, 30.0), (20.0, 20.0), (30.0, 10.0)],
+        []
+    ],
+    type=multipoint_type
+)
+
+write_encoding_files(
+    geometries_wkt, geometries, geometry_type="MultiPoint"
```
rouault, line 142:
```suggestion
    geometries_wkt, geometries, geometry_types="MultiPoint"
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+        [[(30.0, 10.0), (10.0, 30.0), (40.0, 40.0)]],
+        [[(10.0, 10.0), (20.0, 20.0), (10.0, 40.0)],
+         [(40.0, 40.0), (30.0, 30.0), (40.0, 20.0), (30.0, 10.0)]],
+        [],
+    ],
+    type=multilinestring_type
+)
+
+write_encoding_files(
+    geometries_wkt, geometries, geometry_type="MultiLineString"
```
rouault, line 167:
```suggestion
    geometries_wkt, geometries, geometry_types="MultiLineString"
```

test_data/generate_test_data.py
```diff
@@ -0,0 +1,195 @@
...
+        [[[(40.0, 40.0), (20.0, 45.0), (45.0, 30.0), (40.0, 40.0)]],
+         [[(20.0, 35.0), (10.0, 30.0), (10.0, 10.0), (30.0, 5.0), (45.0, 20.0), (20.0, 35.0)],
+          [(30.0, 20.0), (20.0, 15.0), (20.0, 25.0), (30.0, 20.0)]]],
+        [],
+    ],
+    type=multipolygon_type
+)
+
+write_encoding_files(
+    geometries_wkt, geometries, geometry_type="MultiPolygon"
```
rouault, line 194:
```suggestion
    geometries_wkt, geometries, geometry_types="MultiPolygon"
```

rouault (commented):
I've detected that there's a typo on geometry_type  which should be geometry_types . Generated files should be regenerated

jorisvandenbossche:
> I've detected that there's a typo on geometry_type which should be geometry_types . Generated files should be regenerated

Ah yes, that's because I started from https://github.com/opengeospatial/geoparquet/pull/134 which I wrote _before_ we made that name change .. Thanks for the catch!

test_data/generate_test_data.py
jorisvandenbossche: Added some null values!

cholmes (approved):

cholmes:
I'm going to go ahead and merge this in, as it's been open and approved for awhile, and seems like a good edition. Any more work can be done as PR's to main.

tschaub:
@jorisvandenbossche - The `scripts` directory has a `generate_example.py` script and a readme that describes how to install the dependencies for it.  Could we move this new `generate_test_data.py` script to that same directory and update the readme if needed to describe how to install the dependencies for it (or update the `pyproject.toml` or the `poetry.lock` as needed)?  In addition, it would be nice to pull the version identifier out of the `format-specs/schema.json` (as is done in the `generate_example.py` script) to limit the number of places we have to update version numbers [for a release](https://github.com/opengeospatial/geoparquet/wiki/Release-Process).

I'll leave the same comment on #232 (as I think it would be nice to consolidate the examples/test data into one thing that is easier to maintain).

jorisvandenbossche:
> In addition, it would be nice to pull the version identifier out of the `format-specs/schema.json` (as is done in the `generate_example.py` script) to limit the number of places we have to update version numbers [for a release](https://github.com/opengeospatial/geoparquet/wiki/Release-Process).

Yes, I realized last week with the release I suboptimally hardcoded the number here in this PR.

Will try to look at it tomorrow.

Issue #201, opened 2024-03-27
The releases on the repository can be misleading regarding the status of GeoParquet as an OGC Standard


doublebyte1:
The releases on this repository refer to the specification, but looking at the way releases are managed in other OGC repositories (e.g.: [JSON-FG](https://github.com/opengeospatial/ogc-feat-geo-json/releases/tag/v0.2.2), [OGC API Tiles](https://github.com/opengeospatial/ogcapi-features/releases/tag/1.0.1) they can be interpreted as versions of the draft standard. 

In this [PR](https://github.com/opengeospatial/geoparquet/pull/200), the README is updated to reflect that this is a draft standard and the releases are not releases of the standard.

cholmes:
Closing this, as there is now language in the main branch of the repo. Let us know if it needs any more adjustment.

PR #209, opened 2024-04-27
Improves Covering Schema #208

### LINKED ISSUE #208: Covering Schema
The covering schema currently allows any kind of properties, e.g. 

```
{
...
  "covering": {
    "bounding-box": false
  }
...
}
```

is valid according to the schema.

Also, column names can be empty. Is that intentional? In other parts a minimum length of 1 is required.

m-mohr:
Fixes #208


format-specs/schema.json
```diff
@@ -83,31 +85,31 @@
                     "xmin": {
                       "type": "array",
                       "items": [
-                        { "type": "string" },
+                        { "type": "string", "minLength": 1 },
```
rouault, line 88: that doesn't hurt. But isn't that useless given the const value ?

m-mohr: The first array element in the schema can be any text.
The second array element is a const value.

The minLength and const are never checking the same value, so no, it's not useless. ;-)

rouault: oh, right. forgot this was a 2 items array. forgot my comment

jwass (commented):
This looks good to me. Two thoughts:

* Add an invalid example to scripts/test_json_schema.py to show that empty name as the first column name will fail.
* bbox is now required but should additional properties be allowed per your comment in #201? If so, you could add `additionalProperties: false` to covering... but maybe it's fine to leave for now.

m-mohr:
I'll look into the tests.

> bbut should additional properties be allowed per your comment in https://github.com/opengeospatial/
geoparquet/issues/208?

I'm not sure what the intention was here. How it is in this PR it seems most reasonable to me, but maybe there was another intention. I hope others can clarify.

format-specs/schema.json
```diff
@@ -74,7 +74,9 @@
             },
             "covering": {
               "type": "object",
-              "minProperties": 1,
+              "required": [
+                "bbox"
+              ],
```
kylebarron, line 79: I don't think we want to make this required if we want to add other covering types in the future. Or would we remove this `required` when we add the future covering type?

kylebarron: Or is there some way to define the specific key names allowed? E.g. that today we only allow the "bbox" key inside `coverings`, but that in the future other keys may be allowed as well?

m-mohr: Yeah, that's the question. I mean right now it's required, isn't it? Otherwise it's empty or someone puts anything else.

In the future the requirement could be relaxed of course, but for 1.1 it doesn't feel very useful to not require it?!

With regards to your second comment, I'm not sure I understand what you are looking for.

kylebarron: > With regards to your second comment, I'm not sure I understand what you are looking for.

My understanding is that your change means that a `bbox` key is now required. Instead, is there some way in JSON schema to allow an enum of values? So we currently define an enum with a single variant but can expand that in the future?

m-mohr: You could do something like the following to indicate that you expect more fields in the future where having one of them is required:

```
"anyOf": [
  { "required": ["bbox"] }
]
```

It's not really different though, it's just an implicit indication for people reading the schema.

The question really is:

Are `covering: {}` or e.g. `covering: { a: "b" }` something that should be allowed in 1.1? Or do we always expect that in 1.1 there is a bbox member in covering?

kylebarron: I think the expectation is that if the `covering` key exists, then it has a `bbox` member key.

m-mohr: Then the PR is correct. If you want I can update it to the anyOf, I'm pretty neutral on it. What do others think?

kylebarron (approved):

format-specs/schema.json
jwass: Agreed. I think it's fine as-is. Whenever more options are added, the `"required": ["bbox"]` can change to anyOf or oneOf without losing the meaning. And since only the outermost element has additionalProperties set to false, it seems consistent to leave it out here.

cholmes:
What's the status of this one? @jwass - you want to give this an approval? Let's get this one merged in unless anyone has objections.

jwass (approved):

PR #210, opened 2024-04-27
Require two elements in covering elements

m-mohr:
The covering schemas for xmin, xmax etc. allow 0-2 elements in the array, but I think according to the spec it should always be exactly two elements.

### DIFF
--- a/format-specs/schema.json
+++ b/format-specs/schema.json
@@ -86,7 +86,8 @@
                         { "type": "string" },
                         { "const": "xmin" }
                       ],
-                      "additionalItems": false
+                      "minItems": 2,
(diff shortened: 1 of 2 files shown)


jwass (commented):
Looks good to me. Thanks. Maybe add an invalid example to make sure zero or 1 length arrays will fail.

m-mohr:
I've added tests.

kylebarron (approved):

jwass (approved):
Thanks!

cholmes (approved):

jorisvandenbossche (approved):
Thanks!

Issue #115, opened 2022-07-01
Media type


m-mohr:
Hi there,
is there already an agreed media type (e.g. for usage in STAC)?

jorisvandenbossche:
I suppose having a registered MIME type for Parquet itself (the issue you linked) would be a required first step (since a geoparquet file is technically just a parquet file).

m-mohr:
Yes, that's true. But on the other hand if it takes too long to wait for it in Apache, the community will just come up with their own definition. 

Microsoft Planetary Computer is currently using `application/x-parquet`, which I'll also adopt for now.

TomAugspurger:
It would be good to establish a convention that differentiates between parquet and geoparquet. I'm happy to update the media types in the Planetary Computer as needed.

I'm not familiar with the various components of a media type, but COG uses `image/tiff; application=geotiff; profile=cloud-optimized`. If Apache Parquet used `application/apache.parquet` would it be appropriate to use `application/apache.parquet; application=geoparquet`, since geoparquet files are valid parquet files? Maybe it'd be weird to have `application` in the media type twice.

rouault:
> If Apache Parquet used `application/apache.parquet` would it be appropriate to use `application/apache.parquet; application=geoparquet`, since geoparquet files are valid parquet files? Maybe it'd be weird to have `application` in the media type twice.

Not necessarily a helpful answer, but having followed a bit the discussions regarding COG (and not sure we managed to reach an endorsed conclusion), I find that understanding IANA rules for MIME types tends to require a dedicated expertise. For example https://www.rfc-editor.org/rfc/rfc6838.html#page-13 mentions " Media types MAY elect to use one or more media type parameters[...] the names, values, and meanings of any parameters MUST be fully specified when a media type is registered in the standards tree". So I guess that a provision for allowing application=geoparquet should be made at the time application/apache.parquet is registered (unless other rules just ban it or make it already possible...)

m-mohr:
OGC contacted IANA to ask about adding parameters such as profile or application and you can't simply "add" them if the original type is already registered. So you need to discuss that with Apache upfront or otherwise register your own, e.g. `application/geo+apache.parquet` (although I'm not sure they would use a . in the name...)

cholmes:
Moving to beta.2. I think we should also try to get in touch with the core Parquet people and see if we can help them register something, even if it's just `application/vnd.apache.parquet`. 

[...]

PR #212, opened 2024-05-06
Added file extension recommendation

### LINKED ISSUE #211: Clarify recommended file extension
I've seen a couple instances of users saving files with a `.geoparquet` file extension. We should provide a recommendation of whether this is allowed or whether we recommend all files to be saved with a `.parquet` extension. I don't think this is covered yet in the spec.

cholmes:
Adds a recommendation to use `.parquet` for the file extension, as discussed in #211

From what I could tell the core parquet specification does not require a specific file extension, so it didn't seem good to require it.

I did not include 'justification' in the text, like a line about how it's not required by the core specification but that some readers may expect it. I figure we can just refer to the discussion of the issue if people want to understand.

I also stopped short of including the `.geo.parquet` [recommendation](https://github.com/opengeospatial/geoparquet/issues/211#issuecomment-2090650490) from @hobu. I started to write it, but then didn't find any specific recommendation in https://copc.io/, so it seemed maybe a little overspecified? Happy to add it, but was also thinking we could just mention it in a blog post and put up some examples of it. Could also have validators issue a 'warning' and recommend it when people use .geoparquet.

It does feel like the spec could use a bit of reorganization, and it's probably time for a table of contents at the top. I just put this at the very bottom, which does feel quite buried, but the other option was to stick it at the very top. I figured I'd get this in (and maybe #115) and then try a PR that just re-organizes. 

Closes #211

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -289,3 +289,7 @@ The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` elem
 It is reasonable for implementations to require that one of the above `id` elements are present and skip further tests to determine if the CRS is functionally equivalent with OGC:CRS84.
 
 Note: EPSG:4326 and OGC:CRS84 are equivalent with respect to this specification because this specification specifically overrides the coordinate axis order in the `crs` to be longitude-latitude.
+
+## File Extension
+
+It is RECOMMENDED to always use `.parquet` as the file suffix for a GeoParquet file.


format-specs/geoparquet.md
```diff
@@ -289,3 +289,7 @@ The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` elem
 It is reasonable for implementations to require that one of the above `id` elements are present and skip further tests to determine if the CRS is functionally equivalent with OGC:CRS84.
 
 Note: EPSG:4326 and OGC:CRS84 are equivalent with respect to this specification because this specification specifically overrides the coordinate axis order in the `crs` to be longitude-latitude.
+
+## File Extension
+
+It is RECOMMENDED to always use `.parquet` as the file suffix for a GeoParquet file.
```
m-mohr, line 295:
```suggestion
It is RECOMMENDED to use `.parquet` as the file extension for a GeoParquet file. The file extension `.geoparquet` SHOULD NOT be used.
```

m-mohr (approved):

jorisvandenbossche (approved):

format-specs/geoparquet.md
```diff
@@ -289,3 +289,7 @@ The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` elem
 It is reasonable for implementations to require that one of the above `id` elements are present and skip further tests to determine if the CRS is functionally equivalent with OGC:CRS84.
 
 Note: EPSG:4326 and OGC:CRS84 are equivalent with respect to this specification because this specification specifically overrides the coordinate axis order in the `crs` to be longitude-latitude.
+
+## File Extension
+
+It is RECOMMENDED to use `.parquet` as the file extension for a GeoParquet file. The file extension `.geoparquet` SHOULD NOT be used.
```
kylebarron, line 295:
```suggestion
It is RECOMMENDED to use `.parquet` as the file extension for a GeoParquet file. This provides the best interoperability with existing Parquet tools. The file extension `.geoparquet` SHOULD NOT be used.
```

kylebarron (approved):

jorisvandenbossche:
Thanks @cholmes

PR #213, opened 2024-05-06
Added Media Type

cholmes:
Added [application/vnd.apache.parquet](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet) as the media type, as discussed in #115

Like #212 I went with no 'justification' as to why we don't have a 'geo' type, but could add that in.

This is likely not the right organizational structure, but working to get the PR in, and then we can revisit where it goes.


format-specs/geoparquet.md
```diff
@@ -289,3 +289,7 @@ The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` elem
 It is reasonable for implementations to require that one of the above `id` elements are present and skip further tests to determine if the CRS is functionally equivalent with OGC:CRS84.
 
 Note: EPSG:4326 and OGC:CRS84 are equivalent with respect to this specification because this specification specifically overrides the coordinate axis order in the `crs` to be longitude-latitude.
+
+## Media Type
+
+A GeoParquet file ([RFC 8259](https://tools.ietf.org/html/rfc8259)) SHOULD use [`application/vnd.apache.parquet`](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet) as the [Media Type](https://en.wikipedia.org/wiki/Media_type) (previously known as the MIME Type).
```
m-mohr, line 295: Why don't we go for a MUST here?

The RFC link seems to be out of place?

cholmes, line 295:
```suggestion
A GeoParquet file SHOULD use [`application/vnd.apache.parquet`](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet) as the [Media Type](https://en.wikipedia.org/wiki/Media_type) (previously known as the MIME Type).
```

m-mohr: @cholmes You edited my post 😂 Happy that it's not just me who does this by accident ;-)

format-specs/geoparquet.md
```diff
@@ -289,3 +289,7 @@ The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` elem
 It is reasonable for implementations to require that one of the above `id` elements are present and skip further tests to determine if the CRS is functionally equivalent with OGC:CRS84.
 
 Note: EPSG:4326 and OGC:CRS84 are equivalent with respect to this specification because this specification specifically overrides the coordinate axis order in the `crs` to be longitude-latitude.
+
+## Media Type
+
+A GeoParquet file SHOULD use [`application/vnd.apache.parquet`](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet) as the [Media Type](https://en.wikipedia.org/wiki/Media_type) (previously known as the MIME Type).
```
m-mohr, line 295:
```suggestion
If a [media type](https://en.wikipedia.org/wiki/Media_type) (formerly: MIME type) is used, a GeoParquet file MUST use [application/vnd.apache.parquet](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet) as media type.
```

format-specs/geoparquet.md
cholmes: Double fail! I'll just leave this be then. New proposal looks good, I'll add it.

format-specs/geoparquet.md
```diff
@@ -289,3 +289,7 @@ The CRS is likely equivalent to OGC:CRS84 for a GeoParquet file if the `id` elem
 It is reasonable for implementations to require that one of the above `id` elements are present and skip further tests to determine if the CRS is functionally equivalent with OGC:CRS84.
 
 Note: EPSG:4326 and OGC:CRS84 are equivalent with respect to this specification because this specification specifically overrides the coordinate axis order in the `crs` to be longitude-latitude.
+
+## Media Type
+
+If a [media type](https://en.wikipedia.org/wiki/Media_type) (formerly: MIME type) is used, a GeoParquet file MUST use [application/vnd.apache.parquet](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet) as media type.
```
cholmes, line 295:
```suggestion
If a [media type](https://en.wikipedia.org/wiki/Media_type) (formerly: MIME type) is used, a GeoParquet file MUST use [application/vnd.apache.parquet](https://www.iana.org/assignments/media-types/application/vnd.apache.parquet) as the media type.
```

m-mohr (approved):

kylebarron (approved):

jorisvandenbossche (approved):

kylebarron:
CI is currently broken from https://github.com/opengeospatial/geoparquet/pull/214

m-mohr:
CI has been fixed

PR #205, opened 2024-04-18
Added disclosure that 1.0.0 is not yet OGC approved

jatorre:
We need to clarify that OGC has not yet voted to approve the 1.0.0 release and we have been asked to make it clear. Closes #200

PR #214, opened 2024-05-10
Update README.md

ghobona:


kylebarron:
This PR broke CI on this repository. We can discuss whether or not to retain some of these CI checks, but this prevents other PRs from being merged.

jorisvandenbossche:
Also, we had another PR open that was adding a note about this as well: https://github.com/opengeospatial/geoparquet/pull/205. It would have been better to discuss those updates there.

kylebarron:
I made an issue about this in https://github.com/opengeospatial/geoparquet/issues/216

PR #225, opened 2024-05-30
Clarify null/missing crs key

kylebarron:
Added wording to more strongly highlight the difference in semantic meaning between a _missing_ `crs` key and a `crs` key set to `null`.


tschaub (approved):

cholmes (approved):

jorisvandenbossche (approved):
Thanks, that's indeed clearer!

Issue #180, opened 2023-07-27
PROJJSON schema version


kylebarron:
Right now the Parquet json schema links to PROJJSON schema version 0.5:
https://github.com/opengeospatial/geoparquet/blob/ad32ccaf5534a3680e62bf9de3f9837befd9c717/format-specs/schema.json#L39

But PROJJSON is now on version 0.6:
https://proj.org/en/9.2/schemas/v0.6/projjson.schema.json

How do we want to handle the referenced projjson schema version?

tschaub:
I think it makes sense to relax [the `test_json_schema.py` script](https://github.com/opengeospatial/geoparquet/blob/main/scripts/test_json_schema.py) so it doesn't make an assertion about the `$schema` member of the `crs` object.  The important part is that the rest of the `crs` object is compatible with [the projjson schema](https://proj.org/schemas/v0.5/projjson.schema.json) (and that schema doesn't have any requirements about the `$schema` property).

kylebarron:
I agree for #179 specifically we don't need to resolve this, but as a broader question as GeoParquet gets closer to 1.0: how do we want to handle the fact that we're interlinked to another evolving specification? If we go to 1.0, and then PROJJSON makes a breaking change or goes to 1.0 itself, then would we need to release a 2.0 that uses the new version of PROJJSON?

Otherwise users on the latest version of PROJ would be writing Parquet files that don't pass the validator

tschaub:
Yeah, I agree that the broader question is important to address.  Users cannot easily control what version of PROJ is installed, so they cannot easily control what version of PROJJSON is written.  So tools that generate GeoParquet will have a difficult time ensuring that they write out metadata that is compliant with any particular GeoParquet version that depends on a fixed version of PROJJSON.

m-mohr:
It get's even more difficult and an update might be required: https://github.com/OSGeo/PROJ/issues/4088

PR #227, opened 2024-06-07
Updated projjson schema link to 0.7 (and new link)

cholmes:
Opening this PR more to continue the discussion in #180, since if we do want to be updating the schema versions than 1.1 release seems like a good time to.

I'm not sure what is the proper url to put in, I used https://proj.org/en/latest/schemas/v0.7/projjson.schema.json as the 'main' one seems to redirect there: https://proj.org/schemas/v0.7/projjson.schema.json But the 'id' is the latter one. There's additional discussion of this in https://github.com/OSGeo/PROJ/issues/4088

The case for updating the schema is that 0.5 -> 0.7 is mostly (all?) additive, so everything in the past should work.

The case against is that we should do something different than hard code in a version number. Or if we have to since that's how schemas work then we should target a new one.

### DIFF
--- a/format-specs/schema.json
+++ b/format-specs/schema.json
@@ -36,7 +36,7 @@
             "crs": {
               "oneOf": [
                 {
-                  "$ref": "https://proj.org/en/latest/schemas/v0.7/projjson.schema.json"
+                  "$ref": "https://proj.org/schemas/v0.7/projjson.schema.json"
                 },
                 {
                   "type": "null"
--- a/format-specs/schema.json
+++ b/format-specs/schema.json
@@ -36,7 +36,7 @@
             "crs": {
               "oneOf": [
                 {
-                  "$ref": "https://proj.org/schemas/v0.5/projjson.schema.json"
+                  "$ref": "https://proj.org/en/latest/schemas/v0.7/projjson.schema.json"
                 },
                 {
                   "type": "null"


jorisvandenbossche:
> I'm not sure what is the proper url to put in, I used https://proj.org/en/latest/schemas/v0.7/projjson.schema.json as the 'main' one seems to redirect there: https://proj.org/schemas/v0.7/projjson.schema.json But the 'id' is the latter one.

Given that the GDAL documentation about PROJJSON still uses the https://proj.org/schemas/v0.7/projjson.schema.json url, I would follow their lead?

format-specs/schema.json
```diff
@@ -36,7 +36,7 @@
             "crs": {
               "oneOf": [
                 {
-                  "$ref": "https://proj.org/schemas/v0.5/projjson.schema.json"
+                  "$ref": "https://proj.org/en/latest/schemas/v0.7/projjson.schema.json"
```
rouault, line 39:
```suggestion
                  "$ref": "https://proj.org/schemas/v0.7/projjson.schema.json"
```

rouault:
> Given that the GDAL documentation about PROJJSON still uses the https://proj.org/schemas/v0.7/projjson.schema.json url, I would follow their lead?

yes, the redirection to the en/latest/ URL is a ReadTheDocs hosting implementation detail

jorisvandenbossche (approved):

kylebarron (approved):

tschaub:
Should this reference in the example metadata also be changed?
https://github.com/opengeospatial/geoparquet/blob/a684d7a72f950f290e64f057d47932ae03e457c6/examples/example_metadata.json#L32

Issue #228, opened 2024-06-11
Forward compatibility guarantees?


TheNeuralBit:
Hi, I'm looking for clarification on the SemVer policy in the [1.0.0 release notes](https://github.com/opengeospatial/geoparquet/releases/tag/v1.0.0). SemVer is a little unclear to me in the context of a specification for a file format. Specifically, in my case I'm wondering if there are "forward compatibility guarantees" for consumers of GeoParquet metadata, is it guaranteed that a consumer who is only aware of the 1.0.0 spec will be able to consume any 1.x.x metadata? This would be nice for consumers, but it seems like it effectively freezes the format. I'm not sure you can evolve the format under such a constraint. Is this the intention?

As an example of a different approach, the Arrow project uses two different version numbers, one for API compatibility in the client libraries, and one for the format. The API version uses SemVer, and the format version just provides a [forward compatibility guarantee](https://arrow.apache.org/docs/format/Versioning.html#forward-compatibility) - basically older software must be able to consume newer data _or_ be able to detect that it cannot consume it.

paleolimbot:
> older software must be able to consume newer data or be able to detect that it cannot consume it.

I think that characterizes the intent of our semantic versioning here...the `"encoding"` values that we're adding in 1.1 should not cause a problem for existing readers (who should have checked for `"encoding": "WKB"`); and the new bounding box auxiluary column is a feature that can accelerate a scan but can also be safely ignored. I am not sure we've documented that particularly well, though 🙂

TheNeuralBit:
Got it, thanks very much @paleolimbot. Is there a logical place to document this?

paleolimbot:
Perhaps here?

https://github.com/opengeospatial/geoparquet/blob/f8b524f2d5002f485833907d6d00b9174d8f76ae/format-specs/geoparquet.md#L9-L11

jorisvandenbossche:
It would indeed be good to better document this.

One remark about forwards/backwards compatibility is that however we describe it for the format, whether a change is actually compatible still depends on how it is implemented. 

For example adding a _new_encoding, as we did in 1.1, is something we considered as forward compatible and thus fine to do in a 1.1 release (i.e. doesn't require a 2.0), but of course that still depends on the implementation to actually check this value and raise a proper error. 
(it reminded me about the logical types in Parquet, where I think adding a new logical type is also considered as a compatible change, but it turned out that implementations right now would actually crash when encountering a logical type they didn't recognize)

[...]

PR #229, opened 2024-06-13
Add version compatibility documentation

TheNeuralBit:
This PR proposes some version compatibility documentation, as discussed in #228. I'm just aiming to get the conversation started here. I welcome any bike shedding or nits. Or if someone more closely aligned with the project wants to take this over I'm happy to support that as well.

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -8,7 +8,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ## Version and schema
 
-This is version 1.1.0-dev of the GeoParquet specification.  See the [JSON Schema](schema.json) to validate metadata for this version.
+This is version 1.1.0-dev of the GeoParquet specification.  See the [JSON Schema](schema.json) to validate metadata for this version. See [Version Compatibility](#version-compatibility) for details on version compatibility guarantees.
 
 ## Geometry columns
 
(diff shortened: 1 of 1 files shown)


cholmes:
Thanks for the contribution @TheNeuralBit - it looks good! I don't have any particular nits on this, and it'd be a good thing to have in.

cholmes (approved):

jorisvandenbossche:
The examples clarifying the policy look great to me. On the exact terminology of "backwards compatible", I am not entirely sure it is correct or whether it should be "forwards compatible" in context of the format spec.

For example https://arrow.apache.org/docs/dev/format/Versioning.html only uses "backwards compatibility" for library versions, while it used "forward compatibility" for minor versions of the format:

> An increase in the **minor** version of the format version, such as 1.0.0 to 1.1.0, indicates that 1.1.0 contains new features not available in 1.0.0. So long as these features are not used (such as a new data type), forward compatibility is preserved.

Now, as @paleolimbot mentioned in the meeting, maybe the current wording here is fine, as you can say that a backwards compatible addition to the format spec ensures to preserve forward compatibility.

jorisvandenbossche:
As additional context, there is a WIP update for the Parquet format itself to better describe this (https://github.com/apache/parquet-format/pull/258). The current version of the PR defines:

1. Backwards compatible. A file written under an older version of the format
   should be readable under a newer version of the format.

2. Forwards compatible. A file written under a newer version of the format with
   the feature enabled can be read under an older version of the format, but
   some information might be missing or performance might be suboptimal.

3. Forwards incompatible. A file written under a newer version of the format with
   the feature enabled cannot be read under an older version of the format (e.g.
   adding and using a new compression algorithm).

This defines "forward compatibility" a bit different, as we explicitly say that it is fine to add new features as long as old readers can detect that the new feature is being used.

TheNeuralBit:
Thanks for the comments. I think you're right we should talk about forward compatibility here. It's not necessarily true that a backward compatible change is forward compatible. It would be backwards compatible to add a new field that, when specified, changes the interpretation of some other field, but that's not forward compatible (an older reader would get incorrect results by ignoring the new field).

jorisvandenbossche (approved):
Thanks for the update!

cholmes:
Looks great, merging in.

Issue #198, opened 2024-03-11
Antimeridian Crossings and bbox


jwass:
As mentioned in #191 in https://github.com/opengeospatial/geoparquet/pull/191#discussion_r1488344906, geometries that cross the antimeridian will break the current row group filtering mechanics.

```
e.g if you have a geometry [-10,-10,10,10] and a [170,-10,-170,10] (crossing A-M) in a single row group, then the row group stats are going to be [-10,-10,10,10] and thus the geometry crossing A-M will not be selected.
```

Our solution is just say antimeridian crossings are unsupported but we can use this thread to discuss solutions.

rouault:
A potential solution would be to require that for a AM crossing geometry, the bbox.xmin is in the range [0,180], and add +360 to the bbox.xmax so that it is in the range [180,360].   (the x of the geometries themselves could/should still be in the [-180,180] range)
So a  geometry with bbox [170,-10,-170,10] would be actually indexed as [170,-10,190,10] in the bounding box columns

jorisvandenbossche:
@kylebarron is that the same as what you suggested in the meeting two weeks ago (adding 360 to the xmax value)?

While that fixes the issue with the "wrong" rowgroup-level xmax value, how does this work for filtering based on the bbox column itself? Assume you are reading with a bbox filter of [-180, -10, -170, 10]. Using simple x/y min/max comparisons for the bbox, such a filter would also not select the row with the A-M crossing, because that bbox now doesn't overlap with [170,-10,190,10]

rouault:
> Assume you are reading with a bbox filter of [-180, -10, -170, 10]. Using simple x/y min/max comparisons for the bbox, such a filter would also not select the row with the A-M crossing, because that bbox now doesn't overlap with [170,-10,190,10]

True, and I'm pretty sure there is no simple/elegant solution as soon as discontinuities are involved :-)
So if you spatial filter near the anti-meridian, you need to issue your spatial filter twice: one with longitudes near -180 and another time with them shifted of +360.

jorisvandenbossche:
The other option would be to step away from GeoJSON's wording using the "most southwesterly point (min) and most northeasterly point (max)" ? 
And use the actual min/max of the coordinate values instead. For the example we are using, that would give a bbox of something like  [-180, -10, 180, 10]. Of course, that spans a big part of the globe, and makes the bbox very not-specific (making this is selected too often). 
[...]

PR #230, opened 2024-06-17
Updates for the 1.1.0 release

cholmes:
This branch prepares everything for release, in the hopes of being able to discuss the final issues in the next meeting and finalize the release. The remaining issues to discuss and potentially merge includes:

#227 - schema version for projjson
#198 - antimeridian crossing - @jorisvandenbossche had some potential ideas.
#123 - Test data - not sure if Joris had a couple more he wanted to get in.
#229 - version compatibility suggestion that just landed, but seems pretty good.


jorisvandenbossche (approved):

kylebarron (approved):

tschaub:
It looks like there is another version number in the `example_metadata.json` file that needs to be manually updated as part of the release process (alternatively, maybe the `scripts/generate_example.py` script could be updated to generate this file so we have fewer places to update version numbers).

https://github.com/opengeospatial/geoparquet/blob/a684d7a72f950f290e64f057d47932ae03e457c6/examples/example_metadata.json#L138

cholmes:
> It looks like there is another version number in the `example_metadata.json` file that needs to be manually updated as part of the [release process](https://github.com/opengeospatial/geoparquet/wiki/Release-Process) (alternatively, maybe the `scripts/generate_example.py` script could be updated to generate this file so we have fewer places to update version numbers).

Good catch, thanks. Will fix it (and at least mention something in the release wiki).

examples/example_metadata.json
```diff
@@ -135,6 +135,6 @@
       }
     },
     "primary_column": "geometry",
-    "version": "1.1.0-dev"
+    "version": "1.1.0"
```
cholmes, line 138: @tschaub - I think this is right on this branch. Your [link in the comment](https://github.com/opengeospatial/geoparquet/pull/230#issuecomment-2174005239) looks like it goes to the PR for the projjson schema fix - https://github.com/opengeospatial/geoparquet/commit/a684d7a72f950f290e64f057d47932ae03e457c6. I see it right [on the 1.1 release branch](https://github.com/opengeospatial/geoparquet/blob/e61f1516879c2e7f494cb99cab4492144f0f4d91/examples/example_metadata.json#L138). 

Let me know if I'm missing anything.

cholmes:
Ok, I think this is all ready to go. Approvals are now open! Our conclusions on the call were as follows:

https://github.com/opengeospatial/geoparquet/pull/227 - (projjson schema version update) datemerged
https://github.com/opengeospatial/geoparquet/issues/198 - (antimeridian crossing) - we don't feel like we have a great solution all ready, so moving to 1.2 and retaining the official guidance of just saying 'don't do that'.
https://github.com/opengeospatial/geoparquet/issues/123 - Test data - Joris may try to get something in, but there's lots more test data, we're just missing some for 1.0, which we didn't have for 1.0 anyways, so we decided not to block on this (but he may try to get it in soon).
https://github.com/opengeospatial/geoparquet/pull/229 - (version compatibility) merged - with a bit of back and forth on 'forward compatibility'.

And I couldn't find where @tschaub pointed out that the version was off. If reviewers could make sure that there are no 1.1.0-dev versions anywhere on this branch then that'd be much appreciated.

I'm aiming for at least 4 +1's before releasing. When we've got that I'll tag the release and push it out. Thanks all!

examples/example_metadata.json
tschaub: My bad.  Looks right in this change.

cholmes: No prob! Thanks for checking. A +1 on the release is appreciated :)

tschaub:
Looks like it is the `example_metadata_point.json` file that has the wrong version number:
https://github.com/opengeospatial/geoparquet/blob/3048e5f30c7ec42714fb06790f3aff7282a2fd94/examples/example_metadata_point.json#L12

We could:
 1. manually update the version number in this file as well and update the [release process](https://github.com/opengeospatial/geoparquet/wiki/Release-Process) so that others do the same in the future
 2. remove this file (not clear to me that it is used)
 3. update things so that this file is generated from a corresponding `example_point.parquet` file (that would be generated from the `generate_example.py` script and would mean fewer version numbers to maintain)

Any preferences?

cholmes:
Good catch.

I think for now I lean towards #2, as I'm not clear what it's for, and then at some point get better test data.

cholmes:
Oh wait, I see, it's meant to be an example of the geoarrow encoding. So probably shouldn't just remove it. Ideally we'd have an example that is actually in the geoarrow encoding in parquet, not just the metadata.

paleolimbot (approved):

tschaub (approved):

tschaub:
I updated the version in `example_point_metadata.json` and opened #232 to suggest consolidating the example/test data.

PR #234, opened 2024-06-20
Remove requirement that geometry columns must not be a group field

### LINKED ISSUE #233: Geometry nesting requirement is incompatible with "native" encoding
It looks #189 should have also removed the requirement about the geometry column [not being a group](https://github.com/opengeospatial/geoparquet/blob/v1.1.0/format-specs/geoparquet.md?plain=1#L21):

> A geometry MUST NOT be a group field or nested in a group.

This is incompatible with the new native encodings.  The [example in the spec](https://github.com/opengeospatial/geoparquet/blob/v1.1.0/format-specs/geoparquet.md?plain=1#L114-L118) goes against the requirement above:
```
// "point" geometry column as simple field with two child fields for x and y
optional group geometry {
  required double x;
  required double y;
}
```

tschaub:
The new native geometry encodings from #189 are incompatible with the requirement that geometry columns must not be group fields.  This branch suggests removing that requirement.

Fixes #233.

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -18,7 +18,7 @@ See the [encoding](#encoding) section below for more details.
 
 ### Nesting
 
-Geometry columns MUST be at the root of the schema. A geometry MUST NOT be a group field or nested in a group. In practice, this means that when writing to GeoParquet from another format, geometries cannot be contained in complex or nested types such as structs, lists, arrays, or map types.
+Geometry columns MUST be at the root of the schema. In practice, this means that when writing to GeoParquet from another format, geometries cannot be contained in complex or nested types such as structs, lists, arrays, or map types.
 
 ### Repetition
 


cholmes:
Good catch, though do we want to completely throw out the rule? Like we probably still don't want geometries nested in a group? And we could also just say that the only groupings allowed are the native geometries...

I guess with this we'll probably need a 1.1.1?

tschaub:
> Good catch, though do we want to completely throw out the rule? Like we probably still don't want geometries nested in a group? And we could also just say that the only groupings allowed are the native geometries...

I think this might be a case where having more language in the spec is just going to create confusion.

With the change I've proposed, our one requirement related to nesting/groups is that all geometry columns must be at the root of the schema.  This implies (and is the same thing as saying) that geometry columns cannot be nested within other group fields.  So I wouldn't add any additional words related to that.

The encoding section already has a lot of language about how the encoding changes the structure of the geometry field, so I'm not sure more language is needed above.

Basically, I think lots of language invites lots of differing interpretations.  And with fewer words we risk fewer inconsistencies/misinterpretations/bugs etc.

> I guess with this we'll probably need a 1.1.1?

Yeah, I think that makes sense.

cholmes:
Cool, makes sense to me. 

Let's try to get this out soon, but ideally get a few more eyes on this PR.

kylebarron (approved):

jorisvandenbossche:
> I guess with this we'll probably need a 1.1.1?

It's actually something we haven't discussed (or had to discuss) before, but do we want a new version number for something like this? Because it is not fixing something in the spec (in the sense of fixing something in implementations / files), but only fixing a wording in the spec. 
And getting this fixed wording out (in the rendered spec on the website) is definitely good, but do we then actually want files with a version number of "1.1.1" in their metadata?

jorisvandenbossche (approved):
The change itself certainly looks good, thanks for catching that inconsistency!

cholmes:
I guess one option would be to cut a 1.1.1 'release' on github (so there's a tag and the website gets updated), but keep the version the same?

Could even call it 1.1.0.1 on github / spec? Or call it 1.1.0-update or something?

kylebarron:
In Python there's also the idea of "post releases" https://packaging.python.org/en/latest/specifications/version-specifiers/#post-releases

cholmes:
Oh nice, I like that. Good to use some precedent.

cholmes:
Merging in, since no matter what we need this wording in. I'm going to try to run with the 'post release' idea and make a branch to propose it, as I'd like to get this wrapped up soon.

tschaub:
@cholmes - with Semver this would be handled with "build metadata" following the version number - so `1.1.0+P1` would be a version that would follow `1.1.0` and precede `1.2.0` (or `1.1.1` if we decided we wanted that later).  The letter `P` is arbitrary, but we could decide that we want to use `P` build metadata to indicate our post-release fixes.

We could make some changes to the way the website is built to only show the most recent release considering build metadata.

Please don't follow the version naming convention from Python.

tschaub:
Here is a summary of the latest tag and build metadata handling: https://github.com/opengeospatial/geoparquet/pull/236#issuecomment-2183547346

Issue #232, opened 2024-06-18
Consolidate examples and test data


tschaub:
We have a single example Parquet file, a couple JSON metadata files, and an R file in the `examples` directory.  In addition, we have a number of examples Parquet files in the `test_data` directory.  We should consolidate these (likely remove `examples` and point people to `test_data` for examples).

All our example Parquet files should be generated by scripts in the repo.  And if we need to have separate JSON metadata files, these should be generated too (though it feels like we could do without these).  And after we consolidate the examples into a single location, the [release process](https://github.com/opengeospatial/geoparquet/wiki/Release-Process) should be updated to describe how to update the examples for new versions.

PR #236, opened 2024-06-21
Updates for the 1.1.0+p1 release

tschaub:
This includes fixes to the 1.1.0 release (https://github.com/opengeospatial/geoparquet/pull/234).  This is being tagged with build metadata `p1`.  After creating this release, the website needs to be updated to host the most recent version of the spec and schema at the "clean" version while considering build metadata.

~TODO:~
 - ~[ ] update the `test_data` files to include the right version~
 - ~[ ] update the [release process](https://github.com/opengeospatial/geoparquet/wiki/Release-Process) to describe how to update the `test_data`~


scripts/poetry.lock
```diff
@@ -2,21 +2,22 @@
 
 [[package]]
 name = "attrs"
-version = "23.1.0"
+version = "23.2.0"
```
tschaub, line 5: Following [the scripts readme](https://github.com/opengeospatial/geoparquet/blob/main/scripts/README.md#usage), I ran `poetry update`.  This resulted in the lockfile changes below.  If we don't want this to happen, we should update that readme.

cholmes:
> update the test_data files to include the right version

Did they go out in 1.1.0 with the wrong version? Or are you saying they'd get 1.1.0+p1 as the version?

tschaub:
> Did they go out in 1.1.0 with the wrong version? Or are you saying they'd get 1.1.0+p1 as the version?

I was unsure if they were updated as part of the previous release to include the new dev version (`1.2.0-dev`, like with the other [example data](https://github.com/opengeospatial/geoparquet/tree/main/examples)).

I see now that the `test_data` is generated with [a new script](https://github.com/opengeospatial/geoparquet/blob/dced61cac91867b3a1218381786cc43bbf245650/test_data/generate_test_data.py) that has yet another version of the version identifier in it.  This means that the `test_data` will not get updated unless we update the [release process](https://github.com/opengeospatial/geoparquet/wiki/Release-Process) to describe how to do that.

As part of https://github.com/opengeospatial/geoparquet/issues/232 I think it would make sense to make it so the test data is generated with the version number from `format-specs/schema.json` - and ideally put in a place with a readme that describes how to install the required dependencies (like the `scripts` directory).

tschaub:
Here are the website changes to support build metadata in our release tags: https://github.com/geoparquet/geoparquet.github.io/pull/59 (this may need additional changes after we push this tag)

tschaub:
I've pushed the `v1.1.0+p1` tag.  Here is the corresponding release https://github.com/opengeospatial/geoparquet/releases/tag/v1.1.0%2Bp1

The website has been updated to strip the build metadata when extracting the version from a tag (sorting to get the most recent based on [semver](https://semver.org/)).

As mentioned [in the release notes](https://github.com/opengeospatial/geoparquet/releases/tag/v1.1.0%2Bp1), the version to use in the GeoParquet metadata is still `1.1.0`.

The corrected spec is published here (over the old one): https://geoparquet.org/releases/v1.1.0/

Because we didn't delete the old tag, people can still get to the [old](https://github.com/opengeospatial/geoparquet/blob/v1.1.0/format-specs/geoparquet.md) and [new](https://github.com/opengeospatial/geoparquet/blob/v1.1.0%2Bp1/format-specs/geoparquet.md) spec via git.  Validators and other clients should use the published version of [the spec](https://geoparquet.org/releases/v1.1.0/) and [metadata schema](https://geoparquet.org/releases/v1.1.0/schema.json) instead.

jorisvandenbossche:
Thanks @tschaub for streamlining this. Such a workflow with `+p1` metadata sounds good to me.

For this PR, there is no longer an actual relevant diff?

tschaub:
> For this PR, there is no longer an actual relevant diff?

Yeah, the release branch was useful for making release-related changes, tagging the release, and then changing things back to dev versions.  But we could close this without merging now.

PR #237, opened 2024-06-24
Fix typo in multipolygon example

msbarry:
It looks like this line should be `repeated group list` instead of `required group list` in the geoarrow multipolygon example.


tschaub (approved):

jorisvandenbossche (approved):
Thanks for the catch!

PR #248, opened 2024-11-08
Early draft guidance document on general guidance on spatial type implementation

paleolimbot:
Very early draft with plenty of unsubstantiated opinions! Just posting here for feedback 🙂 .

For background, at the last sync we discussed that probably a document like this should exist given the discussions that have taken place in the Parquet and Iceberg PRs. I am not sure this is the actual home for this or if this should be anything more than a blog post, but thought this is the right group of people to discuss this with!


format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+These components are used by major data warehouse vendors like
+[BigQuery](https://cloud.google.com/bigquery),
+[Snowflake](https://snowflake.com), [Databricks](https://databricks.com). Broadly,
+GeoParquet was
+
+## Spatial Types?
+
+By **Spatial Type**, we are referring to a library, specification, or file format
+that intends to represent points, lines, polygons, or combinations thereof and relate
+them to a position on the surface of the earth[^1]. We will also constrain this
```
rouault, line 16:
```suggestion
them to a position on the surface of the Earth[^1]. We will also constrain this
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+"Geometry" [and/or "Geography"](#geometry-and-geography) type, parameterized with `crs`
+([Coordinate Reference System](#coordinate-reference-systems)) as a string, with values
+encoded as [Well-known binary](#well-known-binary) (if your format has a native way
+to store an array of bytes) or [Well-known text](#well-known-text) (otherwise).
+
+In general, the motivation behind the suggestions in this document are to ensure that
+new Spatial Type implementations:
+
+- **Capture producer intent**: Users of your Spatial Type should not have to discard
+  information when converting spatial data from elsewhere. The main thing this means
```
rouault, line 40:
```suggestion
  information when converting spatial data from elsewhere. That means
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+
+For more details and ways to ensure that the type definition is precise and unambiguous,
+keep reading!
+
+## Prior art
+
+GeoParquet is far from the first attempt to formalize the representation of spatial
+data to an existing type system/storage format. A non-exhaustive list of established
+Spatial Type implementations (in no particular order) we will refer to in this document
+include:
```
rouault, line 69:
```suggestion
includes:
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+
+- [Shapely](https://github.com/shapely/shapely)
+- [GEOS](https://libgeos.org)
+- [JTS](https://locationtech.github.io/jts/)
+- [s2geometry](http://s2geometry.io)
+- [Boost Geometry](https://www.boost.org/doc/libs/1_86_0/libs/geometry/doc/html/index.html)
+
+These libraries are used by Geospatial type implementations to handle the details of
+computational geometry; however, thier scope does not include relating those geometries
+to the surface of the earth[^2].
```
rouault, line 110:
```suggestion
to the surface of the Earth[^2].
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+once the context of the original database connection is lost, there is no mechanism
+to obtain the full CRS definition to pass on to another Spatial Type[^3].
+
+[^3]: To mitigate this problem, PostGIS does use generally consistent identifiers
+for common CRSes such that it is usually possible to guess the CRS based on the
+integer identifier.
+
+## Coordinate Reference Systems
+
+A Coordinate Reference System (CRS) is can be conceptualized as a "unit"
```
rouault, line 172:
```suggestion
A Coordinate Reference System (CRS) can be conceptualized as a "unit"
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+for the combinations of x, y, z, and/or m values of which geometries are
+comprised. Just as a value of "5 meters" has physical meaning whereas
+the value "5" does not, geometries with a coordinate reference system have
+physical meaning.
```
rouault, line 176: Speaking about units doesn't give much justice to the concept of CRS. Perhaps quicly mentions that it encodes if the coordinates are geographic (longitude / latitude) or projected, and the underlying reference ellipsoid

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+  library) will be able to choose which format they use to serialize a coordinate
+  reference system when converting to your Spatial Type implementation.
+- Naive producers (e.g., database drivers or file readers that do not and will
+  never link to a third party spatial library of any kind) will have whatever
+  sequence of characters that was stored alongside its Spatial Type implementation.
+
+Because some producers can choose, it is best to provide a reccomendation. A
+slightly opinionated current best option is
+[PROJJSON](https://proj.org/en/9.5/specifications/projjson.html), followed by
+[WKT2:2019](https://datatracker.ietf.org/doc/html/rfc7159). Both serializations
```
rouault, line 203: WKT is not JSON encoded ;-)
```suggestion
[WKT2:2019](https://www.ogc.org/publications/standard/wkt-crs/). Both serializations
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+  link to PROJ, a binding to PROJ, or some other coordinate transformation
+  library), which usually construct some internal "CRS" object using a string.
+- Naive consumers (e.g., database drivers or file readers that do not and will
+  never link to a third party spatial library of any kind) whose job it usually
+  is to pass on whatever information is in your Spatial Type implementation
+  to some other Spatial Type implementation.
+
+### Why can't I just use an "identifier"
+
+Multiple identifiers needed for xy + elevation
```
rouault, line 231: Yes and no. EPSG has codified a number of compound CRS where a EPSG id designates both the horizontal and vertical CRS. The main reason is custom CRS, that is CRS not integrated to an official catalog

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+> ([PostGIS](https://postgis.net/docs/geography.html))
+
+> This type represents data in a round-earth coordinate system. The SQL Server
+> `geography` data type stores ellipsoidal (round-earth) data, such as GPS
+> latitude and longitude coordinates.
+> ([Microsoft SQL Server](https://learn.microsoft.com/en-us/sql/t-sql/spatial-geography/spatial-types-geography))
+
+The main inconsistency between definitions is whether or not the existing implementations
+use spherical formulas to simplify/accellerate calculations or whether strict ellipsoidal calculations are used. The definitions provided by vendors are not even explicit on this
+point in some cases, and we do not reccomend attempting to separate these cases until
```
rouault, line 295:
```suggestion
point in some cases, and we do not recommend attempting to separate these cases until
```

paleolimbot (commented):
Thank you for the feedback! I will circle back to the CRS section this week :slightly_smiling_face:

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+the purposes of this document include:
+
+- [Shapely](https://github.com/shapely/shapely)
+- [GEOS](https://libgeos.org)
+- [JTS](https://locationtech.github.io/jts/)
+- [s2geometry](http://s2geometry.io)
+- [Boost Geometry](https://www.boost.org/doc/libs/1_86_0/libs/geometry/doc/html/index.html)
+
+These libraries are used by Geospatial type implementations to handle the details of
+computational geometry; however, thier scope does not include relating those geometries
```
dmeaux, line 109: Typo:
> however, thier

 should be:

> however, their

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+that include Geography as an option are usually well aware of the problems
+associated with this and provide explicit mechanisms to perform this conversion.
+
+Some Spatial Type implementations have two data types: Geometry and Geography.
+Both Geometry and Geography type implementations interpret coordinate values
+(i.e., any combination of x, y, z, and/or m values) identically; however the
+two types differ with respect to how adjascent coordinates should be connected
+(i.e., edges). Among other things, this affects how area, length, and distance
+between elements are defined. A precise definition of the edge interpretation
+of all Geometry types of which we are aware can found in the widely adopted
```
dmeaux, line 260: > can found

should be:

> can be found

dmeaux (commented):
I think this does a good job explaining to a user why certain decisions were made in the format definition and the context of the state of the art at the time during which those decisions made.

Also, I found a couple of typos that I noted with comments.

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+Because some producers can choose, it is best to provide a reccomendation. A
+slightly opinionated current best option is
+[PROJJSON](https://proj.org/en/9.5/specifications/projjson.html), followed by
+[WKT2:2019](https://www.ogc.org/publications/standard/wkt-crs/). Both serializations
+can be converted losslessly to each other and can represent information that
+cannot be represented by earlier versions of WKT. We reccomend PROJJSON because
+accessing the contents is possible with any off-the-shelf JSON parser (whereas
+to parse WKT2 a specialized parser is required).
+
+Because some producers *cannot* choose, we also reccomend allowing those producers
```
desruisseaux, line 209: reccomend -> recommend

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+
+Multiple identifiers needed for xy + elevation
+
+### Unsolved Coordinate Reference System issues
+
+- Axis Order:
+  https://macwright.com/lonlat/
+  https://erouault.blogspot.com/2024/09/those-concepts-in-geospatial-field-that.html
+- Representing a coordinate reference system in JSON
+
```
desruisseaux, line 239: Well, those issues are partially caused by conflation (not only by open source projects, but also in some OGC groups) of concepts that should have been separated concerns. For example, the confusion around _"pixel is point/area"_ is partially because it has been interpreted as synonymous of _"pixel center/corner"_, while actually they are two independent problems. Likewise, the confusion about axis order is caused by _"CRS orientation"_ being interpreted as aligned with _"display orientation"_, while actually CRS is an *intermediate* step between the data and the visualization.

The above links to blogs explain that there is confusions, but do not really analyse the causes and remedies. Users are left with the impression that the situation is just chaos, while actually the problem is oversimplification. I tried to show which concepts have been conflated, and how to resolve those ambiguities, in an [OGC Testbed report (draft)](http://t20-gimi-ogc-006ced09c39758532bebec9d5488fca26bed7198a2d3b1481b.pages.ogc.org/documents/D013/document.html#Geomatys_contribution_affine_transform). This is not a proposal to link to above Testbed (especially since it still a draft), but rather to avoid this controversial topic since above blogs do not really help in my opinion.

paleolimbot: Agreed that this section needs filling out! I do think we need to explain why the GeoParquet spec chose its axis order language (which is very explicit). Those blogs themselves have good references I was planning to look at but it's a great point that they're intended for a spatial audence (whereas this document is intended for a non-spatial audience).

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+## If you don't read anything else
+
+The simplest possible spatial type implementation that can interoperate with most existing
+Spatial Type implementations without loosing information is a
+"Geometry" [and/or "Geography"](#geometry-and-geography) type, parameterized with `crs`
+([Coordinate Reference System](#coordinate-reference-systems)) as a string, with values
+encoded as [Well-known binary](#well-known-binary) (if your format has a native way
+to store an array of bytes) or [Well-known text](#well-known-text) (otherwise).
+
+In general, the motivation behind the suggestions in this document are to ensure that
```
brendan-ward, line 36:
```suggestion
The suggestions in this document are intended to ensure that
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+new Spatial Type implementations:
+
+- **Capture producer intent**: Users of your Spatial Type should not have to discard
+  information when converting spatial data from elsewhere. That means
+  if your Spatial Type intends to interoperate with any other Spatial Type, you need
+  to provide a mechanism to locate a full Coordinate Reference System definition.
+  For example, [GeoPandas](https://geopandas.org) stores a coordinate reference system
+  at the type level of each array of geometries.
```
brendan-ward, line 44:
```suggestion
  for each column of geometries.
```

(type level is perhaps an implementation detail that makes this just a little harder to follow here as a general example)

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+  to improve on existing spatial standards because many of them have limitations; however,
+  such an improvement whose definition lives in a non-spatial format is more likely to
+  cause confusion and/or misinterpretation than deferring to an existing standard.
+- **Leverage the capabilities of the format**: Every format has a unique set of
+  features and limitations and there is sometimes a tension between deferring to existing
+  standards and ensuring that users of the format can leverage existing tools that work with
+  the format. For example, [GeoJSON](https://geojson.org) encodes geometries as JavaScript
+  objects instead of a text or binary-based format.
+
+For more details and ways to ensure that the type definition is precise and unambiguous,
```
brendan-ward, line 61: This transition sentence seems unnecessary; the next section follows nicely enough without it...

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+
+In addition to established examples of spatial types, several newer libraries are
+in the process of adding spatial type support that we will also refer to in this document:
+
+- [GeoArrow](https://geoarrow.org) (Apache Arrow)
+- [DuckDB Spatial](https://github.com/duckdb/duckdb_spatial) (DuckDB)
+- [Ibis Spatial](https://ibis-project.org/reference/expression-geospatial) (Ibis)
+- [Substrait](https://github.com/substrait-io/substrait/blob/67f93b654e4cc60340a214ba90fe15c5e9de941b/extensions/functions_geometry.yaml)
+
+Some examples of projects that are not considered Geospatial type implementations for
```
brendan-ward, line 99: Above you refer to "Spatial" type and here "Geospatial" type; suggest making those consistent.  Also, the casing is a bit inconsistent (e.g., line 91).

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+For clients reading a table directly, it is possible to look up the type-level CRS
+by querying the `geometry_columns` and `spatial_ref_sys` tables; however, for clients
+executing a complex query, it is necessary to inspect the integer that is sent alongside
+each encoded geometry in the result (which is almost always the exact same value for
+every element in the result). An important consequence of this workaround is that
+once the context of the original database connection is lost, there is no mechanism
+to obtain the full CRS definition to pass on to another Spatial Type[^3].
+
+[^3]: To mitigate this problem, PostGIS does use generally consistent identifiers
+for common CRSes such that it is usually possible to guess the CRS based on the
```
brendan-ward, line 167:
```suggestion
for common CRS values such that it is usually possible to guess the CRS based on the
```

(to avoid awkward pluralization)

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+## Coordinate Reference Systems
+
+A Coordinate Reference System (CRS) can be conceptualized as a "unit"
+for the combinations of x, y, z, and/or m values of which geometries are
+comprised. Just as a value of "5 meters" has physical meaning whereas
+the value "5" does not, geometries with a coordinate reference system have
+physical meaning.
+
+Just as the value "5 meters" must be transformed to be plotted alongside
+the value "5 centimeters", so must geometries be transformed to be plotted
```
brendan-ward, line 179:
```suggestion
the value "5 centimeters", geometries must be transformed to be plotted
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+- Coordinate reference systems are important to keep alongside
+  any geometry in a Spatial Type because after discarding this information
+  their physical meaning is lost.
+- Coordinate reference systems can be serialized to a string and can be constructed
+  from a string.
+
+Producers of Spatial data converting to your Spatial Type implementation
+will fall into one of two categories:
+
+- Spatial aware producers (i.e., producers like GeoPandas that already
```
brendan-ward, line 192:
```suggestion
- CRS aware producers (i.e., producers like GeoPandas that already
```

(to distinguish from spatial data stored in a naive implementation, e.g., as x,y columns; it is CRS that is the focus here)

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+  link to PROJ, a binding to PROJ, or some other coordinate transformation
+  library) will be able to choose which format they use to serialize a coordinate
+  reference system when converting to your Spatial Type implementation.
+- Naive producers (e.g., database drivers or file readers that do not and will
+  never link to a third party spatial library of any kind) will have whatever
+  sequence of characters that was stored alongside its Spatial Type implementation.
+
+Because some producers can choose, it is best to provide a reccomendation. A
```
brendan-ward, line 200:
```suggestion
Because some producers can choose, it is best to provide a recommendation. A
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+can be converted losslessly to each other and can represent information that
+cannot be represented by earlier versions of WKT. We reccomend PROJJSON because
+accessing the contents is possible with any off-the-shelf JSON parser (whereas
+to parse WKT2 a specialized parser is required).
+
+Because some producers *cannot* choose, we also reccomend allowing those producers
+to pass on whatever sequence of characters they have available because this is
+significantly better than dropping the coordinate reference system entirely,
+writing those (potentially out-of-spec) characters to your CRS field anyway,
+or loosing a potential library that otherwise would have supported you. Another
```
brendan-ward, line 213:
```suggestion
or losing a potential library that otherwise would have supported you. Another
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+reason to allow this is that there may be future coordinate reference system
+representations that solve some of the problems we will list below, and mandating
+an existing specification may prevent a producer from writing a better value
+later.
+
+Similarly, there are two types of consumers:
+
+- Spatial aware producers (i.e., producers like GeoPandas that already
```
brendan-ward, line 221:
```suggestion
- CRS aware producers (i.e., producers like GeoPandas that already
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+
+[^5]: It is also not a good idea to blindly label a Geometry as a Geography;
+however, this comes up much less frequently since Spatial Type implementations
+that include Geography as an option are usually well aware of the problems
+associated with this and provide explicit mechanisms to perform this conversion.
+
+Some Spatial Type implementations have two data types: Geometry and Geography.
+Both Geometry and Geography type implementations interpret coordinate values
+(i.e., any combination of x, y, z, and/or m values) identically; however the
+two types differ with respect to how adjascent coordinates should be connected
```
brendan-ward, line 257:
```suggestion
two types differ with respect to how adjacent coordinates should be connected
```

format-specs/spatial-type-guidance.md
```diff
@@ -0,0 +1,328 @@
...
+(i.e., edges). Among other things, this affects how area, length, and distance
+between elements are defined. A precise definition of the edge interpretation
+of all Geometry types of which we are aware can found in the widely adopted
+["OpenGIS Implementation Specification for Geographic information - Simple feature access - Part 1: Common architecture"](https://www.opengeospatial.org/standards/sfa):
+
+> **simple feature** feature with all geometric attributes described piecewise
+> by straight line or planar interpolation between sets of points (Section 4.19).
+
+In contrast, Geography data types are newer and have slightly more variation
+among implementations. Broadly, all of them define adjascent coordinates in
```
brendan-ward, line 267:
```suggestion
among implementations. Broadly, all of them define adjacent coordinates in
```

brendan-ward (commented):
Thanks for working on this @paleolimbot !  A few word-smithing suggestions...

format-specs/spatial-type-guidance.md
desruisseaux: GeoParquet is not very explicit regarding axis order. It suffers from the ambiguities that I tried to describe in the Testbed report, and doesn't comply to OGC directive 14. But we probably don't want to redo this debate here. My proposal would be to just said that there is confusions about axis order, that this problem is not yet resolved in the general case, but we have a status quo in e.g. WKB that works with common cases, and the current GeoParquet specification follows that status quo. But we acknowledge that this is an unsatisfying situation, and a future version may have some changes (e.g., maybe a `permutation` field), depending on how things evolve at OGC.

PR #250, opened 2024-12-04
Add explicit language around edges and ensure available options capture engine behaviours that import/export Parquet

paleolimbot:
This PR is an attempt to improve the language around the "edges" key, to ensure it can capture the intent of all potential producers. We've done more research since the original wording and have some specific examples and use cases where implementations and specifications have struggled. Notably, this PR:

 - Adds a "geodesic" option. I only knew about the spherical approximation because I'd only worked with s2, but it's been rightly pointed out that geodesic edges that consider the ellipsoid are a thing in several major engines.
- Adds examples of producers and consumers that use each edge type
- Adds explicit language around "planar" from simple features access
- Adds practical considerations around importing  a column with an unsupported edge type (basically: you can import them silently if there are no edges; you should force an opt-in or display a prominent warning otherwise).

I've also added a PR to add similar language to GeoArrow https://github.com/geoarrow/geoarrow/pull/79 .

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -216,7 +216,7 @@ do not contain edges (i.e., the column only contains points or empty geometries)
 For columns that contain edges, the error introduced by ignoring the original
 edge interpretation is similar to the error introduced by applying a coordinate
 transformation to vertices (which is usually small but may be large or create
-invalid geometries, particularly if vertices are not closely spaced). Ignroing
+invalid geometries, particularly if vertices are not closely spaced). Ignoring
 the original edge interpretation will silently introduce invalid and/or
 misinterpreted geometries for any edge that crosses the antimeridian (i.e.,
 longitude 180/-180) when translating from `"spherical"` or `"geodesic"` edges
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -181,13 +181,51 @@ It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges`
 
 #### edges
 
-This attribute indicates how to interpret the edges of the geometries: whether the line between two points is a straight cartesian line or the shortest line on the sphere (geodesic line). Available values are:
-- `"planar"`: use a flat cartesian coordinate system.
-- `"spherical"`: use a spherical coordinate system and radius derived from the spheroid defined by the coordinate reference system.
+This attribute describes describing the interpretation of edges between explicitly
+defined vertices.
+
+- `"planar"`: edges will be interpreted following the language of
+  [Simple features access](https://www.opengeospatial.org/standards/sfa):
+
+  > **simple feature** feature with all geometric attributes described piecewise
+  > by straight line or planar interpolation between sets of points (Section 4.19).
+
+- `"spherical"`: Edges follow the shortest distance between vertices approximated
+  as the shortest distance between the vertices on a perfect sphere. This edge
+  interpretation is used by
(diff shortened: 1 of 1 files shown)


format-specs/geoparquet.md
```diff
@@ -181,14 +181,52 @@ It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges`
...
+This attribute describes describing the interpretation of edges between explicitly
+defined vertices.
+
+- `"planar"`: edges will be interpreted following the language of
+  [Simple features access](https://www.opengeospatial.org/standards/sfa):
+
+  > **simple feature** feature with all geometric attributes described piecewise
+  > by straight line or planar interpolation between sets of points (Section 4.19).
+
+- `"spherical"`: Edges follow the shortest distance between vertices approximated
```
jiayuasu, line 193: Can one describe the `perfect sphere` via a geodetic CRS? That way, we can merge `spherical` and `geodesic` together and recommend everyone to look into the `crs` key for interpreting the edge?

rouault: yes, a sphere is described by an ellipsoid with an inverse flattening conventionally set to 0 (it should actually be infinity, but the 0 convention is heavily used)
I believe an issue though (mentionned in some other discussion) is that in some situations some software might apply the formulas for spheres on non-spherical ellipsoids as a simplification avoiding using elliptical integrals

paleolimbot: If we did this we would loose the original CRS information, which is important for ensuring that the position of every single vertex is precisely defined (e.g., a CRS transform from a perfect sphere ellipsoid to UTM, for example, is not the transform we want to invoke on the longitude and latitude of the vertices). The spherical approximation is strictly limited to the (usually small) space between adjacent vertices.

paleolimbot: The difference between the geodesic and spherical are probably always trivial (I can at some point do some math to get a handle on the maximum/average error), but if we want a precise definition, we need it to be there.

jiayuasu: Now I understand. `spherical` mode actually overrides the edge specified in the `crs`. That's where the confusion came from.

I think this is making great progress. I will schedule another meeting among all parties to settle down on the geography type of Iceberg

format-specs/geoparquet.md
```diff
@@ -181,14 +181,52 @@ It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges`
...
+  and [PostGIS](https://postgis.net/docs/geography.html). A common library for
+  interpreting edges in this way is
+  [GeographicLib](https://github.com/geographiclib/geographiclib).
 
 If no value is set, the default value to assume is `"planar"`.
 
 Note if `edges` is `"spherical"` then it is RECOMMENDED that `orientation` is always ensured to be `"counterclockwise"`. If it is not set, it is not clear how polygons should be interpreted within spherical coordinate systems, which can lead to major analytical errors if interpreted incorrectly. In this case, software will typically interpret the rings of a polygon such that it encloses at most half of the sphere (i.e. the smallest polygon of both ways it could be interpreted). But the specification itself does not make any guarantee about this.
 
+If an implementation only has support for a single edge interpretation (e.g.,
+a library with only planar edge support), an column with a different edge type
```
jorisvandenbossche, line 213:
```suggestion
a library with only planar edge support), a column with a different edge type
```

m-mohr:
Can this be closed? Most of this seems to be already in the spec now?

PR #260, opened 2025-05-05
Update python version dependency

cholmes:
For the poetry to not fail


kylebarron:
We might want to update the dependency setup to use uv instead

PR #272, opened 2025-05-05
Update poetry spec for updated Python + Poetry

paleolimbot:
I'm sure there's a way to get poetry to work here (or uv to work without updating the example file)...we were basically using old versions of things and it was no longer running with new versions of things. Happy to defer to another solution here...this is just what I could get to work in 30 minutes or so. I'm sure there's a way to pin the old versions and have it work but we should probably figure out how to make this work with recent dependencies.

@cholmes You may have to just disable the action until I (or somebody more familiar with the Python scripts here) can circle back and fix this!


kylebarron:
I originally added this, but I've since switched all my projects over to `uv`, so I haven't kept up with the latest updates to `poetry`.

TomAugspurger:
Thanks @paleolimbot. I pushed a couple additional fixes:

1. Use pandas to read the CSV file (not a big deal, but it infers the dtypes of the other columns).
2. Apparently, the CSV included here in the repo is slightly different from what geopandas.datasets was fetching. I've updated the expected metadata to match the new values.

TomAugspurger (approved):

jiayuasu (approved):

TomAugspurger:
Thanks @paleolimbot!

Issue #262, opened 2025-05-05
Design CRS/SRID handing for Parquet Raster standard


migurski:
“SRID by itself doesn't mean much if you don't point to a spatial_ref_sys table...”

_Originally posted by @rouault in https://github.com/opengeospatial/geoparquet/pull/259#discussion_r2063286009_

mdsumner:
I have to defer to the expertise here (I know nothing about projjson - other than "it's equivalent to wkt more or less"?) my concern is being limited to a set of crs defined by codes, or assuming those are all epsg. I'm used to full wkt, which I expand from proj strings, or just carry "user input" until real-use time and instantiate with PROJ itself. There are many many variants of laea, stere, lcc, aea,and omerc that have utility (especially in the ocean and at high lats) that won't exist in codes and must be defined uniquely from params. It's probably not too much to require that from practitioners, providing wkt or projjson as input. I also strongly think epsg should never be assumed, always use and encourage full "AUTH:code" form.

migurski:
Looking at [PROJJSON](https://proj.org/en/stable/specifications/projjson.html) because I’ve not used it before, and it feels very heavy to repeat in every raster value. [According to the schema](https://proj.org/en/latest/schemas/v0.2/projjson.schema.json) it doesn't seem possible to use an abbreviated object like `{"id": {"authority": "EPSG", "code": 4326}}`.

I’d support an approach that allows for full PROJJSON but additionally “just publish a table of what PostGIS does somewhere stable and then reference it” for a core list of stable known SRIDs.

jiayuasu:
@migurski @cholmes Parquet file metadata can be used to embed common projjson strings. Maybe we can allow the `srid` value refers to the values stored there?

jiayuasu:
This is how parquet native geo types store projjson strings BTW.

paleolimbot:
Yes, a sort of "local database" in the file metadata could work but extends poorly to the multiple file situation. What if you want to `SELECT * FROM 'some/bucket/*.parquet`? In Parquet implementations we can extract that information before calculating schema equality, but we don't have that ability at the row level (if we want to leverage existing multi file readers, which I think we do).

[...]

PR #276, opened 2025-08-18
Update the SRID definition of the parquet raster standard

jiayuasu:
As detailed in this issue https://github.com/opengeospatial/geoparquet/issues/262 , we now have the projjson definition live on https://spatialreference.org/projjson_index.json. I think it is a good idea to update the Parquet Raster section since we now have a long-live and open definition of SRID.

@cholmes @paleolimbot @migurski should we allow authorities other than EPSG? I saw there are ESRI, .. and so on listed here https://spatialreference.org/projjson_index.json . Some of the ids are more than integers.

### DIFF
--- a/format-specs/parquet-raster.md
+++ b/format-specs/parquet-raster.md
@@ -101,12 +101,14 @@ The allowed URI schemes are:
 CRS is represented as a string value. Writer and reader implementations are
 responsible for serializing and deserializing the CRS, respectively.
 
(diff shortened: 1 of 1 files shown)


migurski (approved):
Agreed with everything above, change looks good and we know we have support for non-EPSG authorities just be reading the stable URL at https://spatialreference.org/projjson_index.json

PR #278, opened 2025-11-03
Initial attempt at 2.0 spec

cholmes:
First draft of a 2.0 specification. 

Changes:
 * Updated with references to Parquet specification
 * Removed Arrow native types
 * Removed covering info (no longer necessary as the native stats are better)
 * Changed edges to be planar and non-planar, and added 'algorithm' with spherical, vincenty, etc.
 
The general philosophy of this is to require the full metadata if you want to be GeoParquet 2.0 compliant, and it must match the Parquet metadata. I could imagine a different philosophy where we remove the duplicate information - GeoParquet would not have crs, geometry types, edges / algorithm, and would provide the additional info - primary column, orientation, epoch.

And another approach I've been contemplating is that if a file adds a geometry or geography in Parquet and does it right we call it 'GeoParquet 2.0 compliant', even if they don't write any additional metadata. I stopped short of trying to figure out how exactly to specify that, but I lean towards going in that direction - use the 'brand name' of GeoParquet to encourage Parquet geospatial implementation. 

I also have been contemplating if it'd be a good idea to do a GeoParquet 1.2, where we leave in covering and arrow types as optional (though maybe deprecated). @maxxen spurred that thinking as I believe he said somewhere it could help with backwards compatibility as duckdb code looks for version 1.x in GeoParquet. But I'm more recently leaning away from that, since older libraries using older versions of parquet readers look to mostly just crash when they encounter a geometry or geography logical type. If they wouldn't crash I'd be inclined for a compatibility mode, but it seems like we'd just be aiming at a narrow slice when people upgraded their parquet reader but are too lazy to change the geospatial handling. So I'm more inclined to just push towards getting all geospatial libraries to implement reading 2.0 / parquet types. But I'm definitely open to arguments as to a 1.2 having a value. It wouldn't be hard to do, but the messaging might be a bit weird.

I have more open questions, but will put them in the follow-up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,8 @@ The 'dev' versions of the spec are available in this repo:
 - [JSON Schema](format-specs/schema.json)
 - [Examples](examples/)
 
+The current dev version is 2.0, which is based on [Parquet Geospatial Logical Types](https://github.com/apache/parquet-format/blob/master/Geospatial.md). The Parquet format now includes core `geometry` and `geography` types and the GeoParquet 2.0 spec provides guidance for geospatial tools to the types, along with some optional metadata not covered in the core Parquet specification.
+
 ## Validating GeoParquet
 
 There are two tools that validate the metadata and the actual data. It is recommended to use one of them to ensure any GeoParquet you produce or are given is completely valid according to the specification:


format-specs/geoparquet.md
```diff
@@ -83,78 +81,13 @@ The optional `epoch` field allows to specify this in case the `crs` field define
...
-required and `optional` fields are permitted (as long as the data itself does
-not contain any nulls).
+WKB geometry columns MUST be stored using the `BYTE_ARRAY` parquet type, with either a `GEOMETRY` or `GEOGRAPHY` logical type, as specified by the Parquet format.
 
 #### Coordinate axis order
 
-The axis order of the coordinates in WKB stored in a GeoParquet follows the de facto standard for axis order in WKB and is therefore always (x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS. This follows the precedent of [GeoPackage](https://geopackage.org), see the [note in their spec](https://www.geopackage.org/spec130/#gpb_spec).
+The axis order of the coordinates in WKB stored in a GeoParquet follows the de facto standard for axis order in WKB and is therefore always (x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS. This is aligned with [Parquet Coordinat Axis Order](https://github.com/apache/parquet-format/blob/apache-parquet-format-2.12.0/Geospatial.md#coordinate-axis-order).
```
cholmes:
```suggestion
The axis order of the coordinates in WKB stored in a GeoParquet follows the de facto standard for axis order in WKB and is therefore always (x, y) where x is easting or longitude and y is northing or latitude. This ordering explicitly overrides the axis order as specified in the CRS. This is aligned with [Parquet Coordinate Axis Order](https://github.com/apache/parquet-format/blob/apache-parquet-format-2.12.0/Geospatial.md#coordinate-axis-order).
```

format-specs/geoparquet.md
```diff
@@ -49,15 +47,14 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
 
 | Field Name     | Type         | Description |
 | -------------- | ------------ | ----------- |
-| encoding       | string       | **REQUIRED.** Name of the geometry encoding format. Currently `"WKB"`, `"point"`, `"linestring"`, `"polygon"`, `"multipoint"`, `"multilinestring"`, and `"multipolygon"` are supported. |
+| encoding       | string       | **REQUIRED.** Name of the geometry encoding format. Only `"WKB"` is supported. |
```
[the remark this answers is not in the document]
paleolimbot: I think keeping `"WKB"` is good. We want to be accurate, but I think we also don't want people to have to rewrite their GeoParquet metadata parsers if we can avoid it. (For the case where the reader is treating GEOMETRY/GEOGRAPHY as a byte array, WKB is still accurate).

.pre-commit-config.yaml
```diff
@@ -6,15 +6,15 @@ default_language_version:
   python: python3
 
 # Optionally both commit and push
-default_stages: [commit]
+default_stages: [pre-commit]
```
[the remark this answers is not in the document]
kylebarron: It's possible that the latest version of pre-commit renamed the stage?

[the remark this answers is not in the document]
cholmes: Ah, possible.

cholmes:
Remaining todo's from call on monday:

* Update the geometry types list to align with parquet native geo types (though we'll likely use the fully spelled out version, not just numbers).
* Clarify bbox - that it should match the total of the row groups. Should also clarify it on 4 vs 6 dimensions, I think we were leaning towards just keeping it 4 dimensions, since the 'for real' answer is to go to the stats.

format-specs/geoparquet.md
```diff
@@ -177,18 +121,30 @@ If no value is set, no assertions are made about winding order or consistency of
 
 Writers are encouraged but not required to set `orientation="counterclockwise"` for portability of the data within the broader ecosystem.
 
-It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges` is `"spherical"` (see below).
+It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges` is `"non-planar"` (see below).
```
jorisvandenbossche:
```suggestion
It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges` is `"spherical"` (see below).
```

Above it was left as spherical?

dwilson1988:
Hello all! I realize I'm a bit late to the party, but as a long time user of GeoParquet, I've been catching up on what's going on with the GeoParquet spec as well as the Geometry and Geography types in native Parquet. 

I was a bit concerned after reading the outcome and discussions about certain decisions like no longer having any mechanism for page-level pushdown and dropping GeoArrow in GeoParquet 2.0. I understand the desire to align with the Parquet spec, but I think that may be a mistake. I was honestly pretty surprised that WKB was settled on as the encoding for Parquet when there are quite a few benefits of using GeoArrow over WKB. I understand the choice to drop support for page-level GeospatialStatistics, but unfortunately between decisions made on the Geometry types in Parquet as well as the current choice to drop GeoArrow AND covering columns leaves the format partially hamstrung.

## Coverings and GeoArrow

Without covering columns (agree they are a bit of a workaround and inconvenient for users) OR GeoArrow, query performance in many real world cases is bound to be slower. Let me be clear on why:

Coverings not only place min/max values in the Statistics for each ColumnChunk, but also in each Page. Since there are typically many Pages to a ColumnChunk, having statistics on the Page level (ColumnIndexes) allows for much more granular query ability. I think we all agree Coverings aren't the best solution moving forward, but with WKB they are CURRENTLY the only way to do filter pushdown at the page level.

GeoArrow, on the other hand, supports page level filter pushdown WITHOUT adding a Covering. This is due to the Dremel encoding that Parquet uses on the x/y/[z]/[m] leaf columns - they are stored essentially as a flattened array of values, which means both Page level ColumnIndexes as well as efficient reads and decoding. With more of the data community settling on Arrow, I'm honestly a little surprised that didn't end up as the Parquet standard for Geometry/Geography types as well.

We at SeerAI have been leveraging GeoParquet with great success for years now and have a fast query engine for geospatial queries. In particular, we are able to serve live map queries on datasets as large as the Overture Building Footprints. Our GeoParquet reader is able to handle GeoParquet 1.1 (both WKB and GeoArrow) and as we were recently adding the native Parquet types, that's when we noticed the movement on the two specs.

## Benchmarks

So is this just hypothetical? Does GeoArrow make a difference? Do page level statistics matter? Yes, but with come caveats.

Of course, it depends on the situation - the smaller the row group, the less page size matters, but with larger row groups (and better compression) the differences between the formats can be stark. I think we're in a unique position to be able to run some apples to apples comparisons on the various formats.

Experiment setup:
- My Laptop: 
   - CPU: Intel(R) Core(TM) i9-10980HK CPU @ 2.40GHz
   - OS: Ubuntu 20.04
- Use one file from Overture Maps Building Footprints (~10M rows)
- Use GDAL (since that's what many people will use) to rewrite the file with different options
  - Row Group Sizes: 16k, 66k, 150k
  - Encodings: WKB, Native, GeoArrow
  - Covering or No Covering
  - No change in sort order
  - ZStd Compression
- Perform a geospatial intersect query with each format and record:
  - File Size
  - Query Time

Some caveats here: This is using our proprietary query engine, but no special optimizations were made for specific files. We attempted to use every feature at our disposal to make each format perform as good as we can. Some readers may see different results if they are not using all available query optimization techniques. The raw numbers don't matter as much as the relative numbers.

The original file used was 2025-02-19.0/theme=buildings/type=building/part-00028-bc5e53eb-b1f2-4193-9ba4-84e6c7ca7995-c000.zstd.parquet since I had a local copy of it. 

Here are the results:

- **Total Feature Count**: `10,404,360`
- **BBox for Query**: `[-104.778104, 39.597333, -104.750642, 39.614196]`
- **Features Returned By Query**: `2804`

| Encoding                      | Row Group Size | Covering? | File Size (GB) | Query Time (s) | 
| --------------------------|-------------------|------------|---------------|--------------------|
| part-00028-bc5e53eb-b1f2-4193-9ba4-84e6c7ca7995-c000.zstd.parquet  | Varied                 | YES           | 1.2369          |  1.6                       | 
| GeoParquet WKB         | 150000                | YES           | 1.3183           |  12.0                    | 
| GeoParquet WKB         | 65535 (default)   | YES           | 1.3154          |   6.7                      | 
| GeoParquet WKB         | 16000                  | YES           | 1.3517         |  0.5                        | 
| GeoParquet GeoArrow | 150000               | NO            | 1.0850         | 1.5                        | 
| GeoParquet GeoArrow | 65535 (default)  | NO            | 1.1551         | 1.2                        | 
| GeoParquet GeoArrow | 16000                | NO            | 1.3662        | 0.87                       | 
| Parquet GEOMETRY      | 150000              | NO             | 1.1095        |  12.2                      | 
| Parquet GEOMETRY      | 65535 (default) | NO             | 1.1053         |  6.7                       | 
| Parquet GEOMETRY      | 16000                | NO            | 1.1464         | 0.6                        |


Some qualitative observations:
- GeoArrow shows consistently fast performance across all row group sizes as well as small file sizes (though on average slightly larger than Native)
- WKB with coverings with SMALL ROW GROUPS is really fast since the smaller the row group, the less ColumnIndexes matter since we can quickly rule out ColumnChunks at a time. 
- Native also works well with small row groups
- The Original encoding had varied sized row groups and was made more optimized from the start. The typical row group sizes were around 100k, yet it performs well and has smaller file sizes than the GDAL written WKB. I'm not entirely sure what accounts for the differences

One thing is clear to me from these results: GeoArrow represents consistent performance without much focus on the row group size. For Native Parquet vs GeoParquet WKB, the performance curve on query time is nearly identical, but with row group optimization (not just size), Coverings allow you to squeeze out additional performance. Native Parquet had the most consistently small file sizes, GeoArrow has the smallest. 

Instead GeoParquet 2.0 coalescing on the Parquet standard, I believe the GeoParquet 2.0 spec should lead the charge on what should come next in the Parquet standard. The whole community would benefit from Geospatial support in Parquet being as good as possible. If Geoparquet 2.0 is essentially Parquet, but with a bit of extra metadata, is it buying us anything? What if it were a testbed for the next generation? 

Anyway, that's where my head is at and I apologize for not joining the conversation sooner as I realize this is a lot to digest and we're late in the game. 

Thoughts?

dwilson1988:
Thanks for the response, Dewey! 

To be clear, I would not advocate for keeping covering columns in the GeoParquet 2.0 - I'm 100% aligned on that being the right call. The removal of GeoArrow AND Coverings though without an alternative, does make potential query performance worse on 2.0 than 1.1, which is definitely worth consideration. 

I was unaware of a way to add extra content to the footer aside from KeyValue metadata - is this spec compliant? If so, I do think the ability to write a proper spatial index would help. One alternative could be a spatial page index in each of the `ColumnChunk` KeyValue metadata since there's typically a low number of columns, this probably wouldn't be overly cumbersome to the files. 

Anyway, I'd just like to see 2.0 not be a performance step back from 1.1 and while the community can of course still create 1.1 spec files, justifying the development time to support an older format if not already supported can be difficult. The good news is 1.1 is supported by the majority of open source readers. Anything we can do to improve 2.0 before moving forward with this proposal? I'm happy to put some work in and join in on more conversations. GeoParquet is one of the best improvements in Geospatial in recent memory - I'd love to see it become as good as it can possibly be.

kylebarron:
I think you mean @jorisvandenbossche

BlakeOrth:
@paleolimbot @cholmes 

> It is! https://datafusion.apache.org/blog/2025/07/14/user-defined-parquet-indexes/ . This and the column chunk approach would need some prototyping and low-level engine integration that I don't currently have the bandwidth to do, but if you're interested feel free to open a dedicated issue and we can discuss/iterate! (There are a number of files at https://geoarrow.org/data that are already available in native/wkb/new types if you are looking for more targets to benchmark).

I've reviewed this blog post (thanks a bunch for that link, I had no idea this was allowed) and I think I can allocate some of my time to build some prototypes using `arrow-rs/parquet` and likely `datafusion` as I'm already at least somewhat familiar with both packages. I don't want to derail and muddy up the discussion here more than necessary; what would be the preferred way to collaborate on these prototypes?

m-mohr:
I'm wondering, whether the arrow types could be an extension to GeoParquet 2.0? We'd want the native Parquet types for broad compatibility, but couldn't we define a "more performant" and additive extension that specifies how a column could be added that is based on arrow types? Yes, you'd store the spatial information twice, but as it's a columnar format, people could then just read what they need.

A similar approach could be taken for the bbox / covering. It is purely additive, but well-specified so that for those usecases we don't fragment within the geo domain. It's follows the idea of [Vecorel](https://vecorel.org).

- GeoParquet 2.0: Defines what is in this PR, use of the native parquet types and the additional metadata.
- GeoParquet [2.0] - Arrow Extension: Adding the arrow encoding from 1.1
- GeoParquet [2.0] - Bounding Box Extension: Adding the covering/bbox from 1.1

m-mohr:
It is niche, but would provide a given way how to solve these cases instead of everyone having to invent their own way (both for arrow types and bbox/covering) or falling back to 1.1.

format-specs/geoparquet.md
```diff
@@ -69,106 +66,55 @@ If the `crs` key does not exist, all coordinates in the geometries MUST use long
...
+In a dynamic CRS, coordinates of a point on the surface of the Earth may change with time. To be unambiguous, the coordinates must always be qualified with the epoch at which they are valid.
 
-```
-// "point" geometry column as simple field with two child fields for x and y
-optional group geometry {
-  required double x;
-  required double y;
-}
-```
+The optional `epoch` field allows to specify this in case the `crs` field defines a dynamic CRS. The coordinate epoch is expressed as a decimal year (e.g. `2021.47`). Currently, this specification only supports an epoch per column (and not per geometry).
```
paleolimbot: I'm OK with keeping this, although it is probably ignored everywhere?

format-specs/geoparquet.md
```diff
@@ -183,54 +129,38 @@ It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges`
 
 This attribute indicates how to interpret the edges of the geometries: whether the line between two points is a straight cartesian line or the shortest line on the sphere (geodesic line). Available values are:
 - `"planar"`: use a flat cartesian coordinate system.
-- `"spherical"`: use a spherical coordinate system and radius derived from the spheroid defined by the coordinate reference system.
+- `"spherical"`: use a spherical coordinate system and radius derived from the spheroid defined by the coordinate reference system. The `algorithm` field further specifies how the edge interpolation is done.
```
paleolimbot: I suppose our initial definition was vague enough that the extra metadata key doesn't hurt. In theory this would be a problem with `{"edges": "spherical", "algorithm": "karney"}`, where an engine would have to be careful not to ignore the `"algorithm": "karney"`. In practice ignoring the algorithm is probably not a big deal (small difference compared to spherical/planar).

format-specs/geoparquet.md
```diff
@@ -183,54 +129,38 @@ It is RECOMMENDED to always set the orientation (to counterclockwise) if `edges`
...
 
-For non-geographic coordinate reference systems, the items in the bbox are minimum values for each dimension followed by maximum values for each dimension. For example, given geometries that have coordinates with two dimensions, the bbox would have the form `[<xmin>, <ymin>, <xmax>, <ymax>]`. For three dimensions, the bbox would have the form `[<xmin>, <ymin>, <zmin>, <xmax>, <ymax>, <zmax>]`.
+For non-geographic coordinate reference systems, the items in the bbox are minimum values for each dimension followed by maximum values for each dimension. For example:
+- XY (two dimensions): `[<xmin>, <ymin>, <xmax>, <ymax>]`
+- XYZ (three dimensions): `[<xmin>, <ymin>, <zmin>, <xmax>, <ymax>, <zmax>]`
+- XYM (two dimensions with measure): `[<xmin>, <ymin>, <mmin>, <xmax>, <ymax>, <mmax>]`
+- XYZM (three dimensions with measure): `[<xmin>, <ymin>, <zmin>, <mmin>, <xmax>, <ymax>, <zmax>, <mmax>]`
```
paleolimbot:
```suggestion
- XYZM (three dimensions with measure): `[<xmin>, <ymin>, <zmin>, <mmin>, <xmax>, <ymax>, <zmax>, <mmax>]`

It is not currently possible to specify M bounds without Z bounds using a GeoParquet metadata bbox: in this case, producers may produce an XY bounding box and omit M bounds. M bounds are typically encoded in Parquet statistics for consumers that benefit from this information.
```

...I think keeping the 3D bbox definition as is makes things less ambiguous / less likely to break prior assumptions about bboxes. I have a feeling all one person who will ever produce a 100% XYM GeoParquet file can make do with the embedded Parquet statistics.

format-specs/geoparquet.md
```diff
@@ -49,15 +47,14 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
...
 | geometry_types | \[string]    | **REQUIRED.** The geometry types of all geometries, or an empty array if they are not known. |
 | crs            | object\|null | [PROJJSON](https://proj.org/specifications/projjson.html) object representing the Coordinate Reference System (CRS) of the geometry. If the field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means the data in this column must be stored in longitude, latitude based on the WGS84 datum. |
 | orientation    | string       | Winding order of exterior ring of polygons. If present must be `"counterclockwise"`; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
 | edges          | string       | Name of the coordinate system for the edges. Must be one of `"planar"` or `"spherical"`. The default value is `"planar"`. |
+| algorithm      | string       | Describes the edge interpolation algorithm for non-planar edge interpolation. Must be one of `spherical`, `vincenty`, `thomas`, `andoyer`, `karney`. The default value is `spherical`.
```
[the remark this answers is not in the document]
cholmes: Sure, that works for me.

PR #280, opened 2026-01-28
Update example file for GeoParquet 2.0

paleolimbot:
Just a draft until #278 merges! This PR updates the example file to (1) write Parquet GEOMETRY as the storage type and (2) a few nits I noticed in the file.


cholmes:
Just tried it out - thanks for doing this. It looks to be working, though geoparquet-io didn't pick up a geometry type from the Parquet metadata. If you set it let me know and I can fix my code, if you didn't set that it would be good to.

I'd say feel free to merge. I do think we should have a wider set of example data, showing projections and geography at the very least. And our test_data directory still includes arrow types, so we should get rid of those / update that script to use the stuff here. If I find time I can likely do that, but I'm slammed for the next bit.

cholmes:
@copilot resolve the merge conflicts in this pull request

cholmes (approved):
Looks great, going to merge. I may take a crack at more extensive test date, that exercises more of the fields, but will do that in a separate PR.

PR #284, opened 2026-06-01
GeoParquet 2.0: allow <authority>:<code> CRS in addition to inline PROJJSON

cholmes:
Permits writers to emit the Parquet `crs` property as either inline PROJJSON (the canonical, preferred form) or an `<authority>:<code>` string (e.g. `EPSG:4326`, `OGC:CRS84`). The other Parquet-core forms (`srid:<id>`, `projjson:<key>`) remain disallowed for writers, but readers SHOULD continue to tolerate them for interoperability with the broader Parquet geospatial ecosystem.

The GeoParquet column-metadata `crs` field is widened to accept the same `<authority>:<code>` form, and the two CRS fields (Parquet logical-type `crs` and GeoParquet column-metadata `crs`) MUST mirror each other exactly in both form and value.

### Why

The current 2.0 draft requires inline PROJJSON for any non-default CRS, which forces every writer to ship a PROJJSON generator (pyproj or equivalent) even to emit `EPSG:4326`. That's a real burden for pure-SQL pipelines, lightweight browser-side writers, and high-file-count workloads. Apache Parquet's own geospatial spec lists `<authority>:<code>` as a first-class option alongside PROJJSON; constraining writers to those two forms (and rejecting the more ambiguous `srid:<id>` and indirect `projjson:<key>` forms) gives us alignment with Parquet core without exposing readers to four parsing paths.

PROJJSON remains the **preferred** writer form per spec language; `<authority>:<code>` is a permitted shortcut.

### Summary of changes

- `format-specs/schema.json`: `crs` field now accepts `object` (PROJJSON), `string` (matching `^[A-Za-z][A-Za-z0-9_-]*:[A-Za-z0-9_.-]+$`), or `null`.
  - Column-metadata `crs` row updated (`object|string|null`).
  - §`crs` subsection rewritten to enumerate the two permitted forms and the writer preference.
  - §`crs` Parquet property subsection rewritten with explicit writer MUST / MUST NOT / SHOULD / MAY rules, the reader MUST + SHOULD-tolerate rules, and a mirroring table for the GeoParquet/Parquet `crs` correspondence.
  - §OGC:CRS84 details: added a sentence noting that authority-code `OGC:CRS84` and `EPSG:4326` are likewise equivalent to OGC:CRS84.
- `scripts/test_json_schema.py`: new positive cases (`crs_authority_code_*`) for the four common authority codes; new negative cases (`crs_string_*`) for malformed strings; the previous `crs_string` invalid case (for `"EPSG:4326"`) is now valid.

### Mirroring table (from the new spec text)

| Parquet logical-type `crs`            | GeoParquet column-metadata `crs`        | Meaning                                  |
| ------------------------------------- | --------------------------------------- | ---------------------------------------- |
| absent (Parquet default)              | absent                                  | OGC:CRS84                                |
| inline PROJJSON object                | the same inline PROJJSON object         | CRS fully described in metadata          |
| `<authority>:<code>` string           | the same `<authority>:<code>` string    | CRS identified by authority code         |
| absent (writer signaling "no CRS")    | `null`                                  | CRS explicitly undefined / unknown       |

### Open / followup

The treatment of `null` (explicit "no CRS / unknown") in 2.0 is preserved here for 1.x compatibility but is awkward: Parquet core has no native way to express "no CRS." A followup discussion may propose removing `null` from 2.0; not in scope for this PR.

### Test results

```
57 passed in 0.5s  (test_json_schema.py)
72 passed in 12.2s (full scripts/ test suite)
```

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -49,7 +49,7 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
 | -------------- | ------------ | ----------- |
 | encoding       | string       | **REQUIRED.** Name of the geometry encoding format. Only `"WKB"` is supported. |
 | geometry_types | \[string]    | **REQUIRED.** The geometry types of all geometries, or an empty array if they are not known. |
-| crs            | object\|null | [PROJJSON](https://proj.org/specifications/projjson.html) object representing the Coordinate Reference System (CRS) of the geometry. If the field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means the data in this column must be stored in longitude, latitude based on the WGS84 datum. |
+| crs            | object\|string\|null | [PROJJSON](https://proj.org/specifications/projjson.html) object, or an `<authority>:<code>` string (e.g. `"EPSG:4326"`), representing the Coordinate Reference System (CRS) of the geometry. The value MUST equal the Parquet logical-type `crs` property in the same form. If the field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means the data in this column must be stored in longitude, latitude based on the WGS84 datum. |
 | orientation    | string       | Winding order of exterior ring of polygons. If present must be `"counterclockwise"`; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
 | edges          | string       | Describes how to interpret the edges of the geometries. Must be one of `planar`, `spherical`, `vincenty`, `thomas`, `andoyer`, `karney`. The default value is `planar`.
 | bbox           | \[number]    | Bounding Box of the geometries in the file, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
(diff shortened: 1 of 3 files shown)


format-specs/schema.json
```diff
@@ -38,6 +38,11 @@
                 {
                   "$ref": "https://proj.org/schemas/v0.7/projjson.schema.json"
                 },
+                {
+                  "type": "string",
+                  "description": "An <authority>:<code> CRS reference (e.g. 'OGC:CRS84', 'EPSG:4326', 'EPSG:3857', 'IGNF:ATI').",
+                  "pattern": "^[A-Za-z][A-Za-z0-9_-]*:[A-Za-z0-9_.-]+$"
```
m-mohr, line 44: Is the pattern backed by any specification? It seems rather arbitrary.

format-specs/geoparquet.md
```diff
@@ -59,27 +59,49 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
...
+GeoParquet 2.0 is more restrictive:
+
+- A GeoParquet 2.0 writer MUST use one of:
+  - inline PROJJSON, or
+  - an `<authority>:<code>` string.
+- A GeoParquet 2.0 writer MUST NOT use `srid:<identifier>` (no authority context; ambiguous across systems) or `projjson:<key_name>` (indirect reference via another metadata key; no ergonomic benefit over inline PROJJSON).
+- A GeoParquet 2.0 writer SHOULD prefer inline PROJJSON. It is self-describing, requires no external CRS registry, and matches the GeoParquet 1.x convention.
+- A GeoParquet 2.0 writer MAY use `<authority>:<code>` when:
+  - the CRS is precisely identified by a well-known authority entry, AND
+  - the writing environment lacks a PROJJSON generator (e.g., pure-SQL pipelines, lightweight browser-side writers), OR
```
m-mohr, line 88: This moves the requirement to ship a heavy database from the writer to the reader. I'd assume readers occur more often, so this somehow worsens the situation IMHO.

But if native GEOMETRY already does this authority:code syntax, I guess the ship has sailed?

format-specs/geoparquet.md
```diff
@@ -59,27 +59,49 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
...
+  - reducing per-file metadata size is desirable (e.g., very many very small files).
```
m-mohr, line 89: Is this really a usecase?

format-specs/schema.json
rouault: The only CRS related spec regarding names of authorities or codes is WKT CRS, which allows pretty much everything (cf "wkt Latin1 text", https://docs.ogc.org/is/18-010r11/18-010r11.pdf page 15, which would allow colon itself in the authority or code, although that would be a bad idea).  The proposed pattern is indeed a bit restrictive, but it should capture all real world CRS codes, at least the ones in proj.db.
It has the advantage to exclude things like "EPSG:4326+3855" that PROJ understands (WGS 84 with EGM2008 height), but that might not be universally understood besides it.

cholmes: > Is the pattern backed by any specification? It seems rather arbitrary.

No, it was just an attempt to capture the ones that are actually valid in spatialreference.org right now. I'm totally fine to relax it, like if we wanted to enable "EPSG:4326+3855". The idea on being restrictive was to help people catch spelling errors / typos.

format-specs/geoparquet.md
cholmes: > But if native GEOMETRY already does this authority:code syntax, I guess the ship has sailed?

Yeah, that was my main thinking. I went back and forth for awhile - the geoparquet 'way' is clearly to ship with all the info needed. But with 2.0 we want to enable the adoption of parquet geometry / geography types, and they have that option. They have a weird srid:34 option which just seems bad, so we don't endorse that. And they have a proj json reference in a file, which is hard to use, so we don't recommend that. 

I didn't end that strongly on adding this vs. sticking to projjson, so could be convinced the other way. But it seems like most tools will want to implement the support for what the Parquet spec says, so they'd have to do this anyways. 

It does make things a bit easier for writers, to not have to fully create projjson. But does make it more difficult for readers. I think part of my hope is that we could lean on spatialreference.org to not make people ship the full database. We could make that pattern a bit more explicit.

format-specs/geoparquet.md
cholmes:
```suggestion
```

No, not really. Should have read what my claude came up with a bit more closely.

format-specs/geoparquet.md
m-mohr: In this case, wouldn't it scale better to only have the writers read from spatialreference.org once instead of a client that may even do thousands of requests per second or so?

The question to me is: What is the intent behind having GeoParquet 2 on top of the native types?
Is it just mirroring the native types and add a bit on top or is it meant to be a best practice on top? In the second case, I feel like I'd go with only projjson in the spec as it seems better to embed it during writing instead of requesting it each time a client opens it. If it's the first option, we could go either way.

format-specs/schema.json
m-mohr: Interesting. The + case seems too difficult for me.

A slightly simpler version, if we want:
```suggestion
                  "pattern": "^[\w.-]+:[\w.-]+$"
```

format-specs/geoparquet.md
```diff
@@ -59,27 +59,49 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
...
+
+The GeoParquet column-metadata `crs` field and the Parquet logical-type `crs` property MUST mirror each other exactly:
+
+| Parquet logical-type `crs`              | GeoParquet column-metadata `crs`           | Meaning                                       |
+| --------------------------------------- | ------------------------------------------ | --------------------------------------------- |
+| absent (Parquet default)                | absent                                     | OGC:CRS84                                     |
+| inline PROJJSON object                  | the same inline PROJJSON object            | CRS fully described in metadata               |
+| `<authority>:<code>` string             | the same `<authority>:<code>` string       | CRS identified by authority code              |
+| absent (writer signaling "no CRS")      | `null`                                     | CRS explicitly undefined / unknown            |
```
jorisvandenbossche, line 102: The "absent" in the first column here is confusing, because it is the same as the first row, but with different mappings to GeoParquet?

jorisvandenbossche:
> which forces every writer to ship a PROJJSON generator (pyproj or equivalent) even to emit `EPSG:4326`

FWIW, I would say this part of the rationale is not really true. If your writer only wants to emit `EPSG:4326`, you can easily hardcode the PROJJSON for that instead of hardcoding the `authority:code` string (like we list the `OGC:CRS84` value in the spec as detail)

format-specs/geoparquet.md
paleolimbot: It's an admittedly niche use case, but this shows up in Iceberg appends (where 4KB of PROJJSON in the always uncompressed footer can be >50% of the file).

cholmes:
We had lots of discussion of this on the community call on June 1st. No one feels that strongly, and realistically what we do in the spec isn't going to have a huge effect on the ecosystem - indeed we hope that readers will implement all the recommendations in the parquet spec. 

It did come up that with spatial reference.org clients would not necessarily have to ship a full projection database, and that we could lay out that path a bit in our spec, to explain that they can call spatial reference and look up the projjson. But if that's the route then it seems better to just have an explicit crs with the link to the projjson, with a new prefix, as the parquet spec allows anything in that field. 

There was a slight lean in the group towards just not allowing it and sticking just with projjson, just to stay a bit more backwards compatible and to nudge the ecosystem towards in-lining projjson.

cholmes:
closing in favor of #286, which came from discussions of the core group.

PR #285, opened 2026-06-01
GeoParquet 2.0: discuss removing null from the crs field

cholmes:
> **Stacked on #284.** This PR is based on the `crs-representation-2.0` branch from #284 and depends on it. Once #284 lands (or is rebased), this can be retargeted to `main`. Reviewing the diff against #284's branch shows only the null-specific changes.

## Goal

Open discussion on removing `null` as a permitted value of the GeoParquet column-metadata `crs` field in GeoParquet 2.0.

## Motivation

In GeoParquet 1.x, `crs: null` means "CRS is undefined or unknown" — distinct from "absent" (which defaults to OGC:CRS84).

In GeoParquet 2.0, the Parquet `GEOMETRY`/`GEOGRAPHY` logical-type `crs` property is the source of truth. The Parquet core specification has **no native way to express "no CRS / unknown CRS"** — absence defaults to OGC:CRS84. PR #284 (which this PR stacks on) preserves 1.x semantics by saying: if GeoParquet `crs: null`, the writer omits the Parquet `crs`, and the GeoParquet `null` is authoritative and overrides the Parquet default-to-OGC:CRS84. This works but adds a corner case to every reader's logic and a special row to the mirroring table.

## Proposed change

In GeoParquet 2.0:

- The column-metadata `crs` field MUST NOT be `null`.
- Permitted values are: a PROJJSON object, an `<authority>:<code>` string, or absent (defaulting to OGC:CRS84) — mirroring the Parquet logical-type `crs` exactly.

Writers that genuinely do not know the CRS of their data should continue to use GeoParquet 1.x.

### For disallowing `null`

- **Removes a special case.** Readers and writers each get one fewer branch. The mirroring rule between Parquet `crs` and GeoParquet `crs` becomes a literal equality check (and the mirroring table loses a row).
- **Honest about Parquet's model.** GeoParquet 2.0 is built on the Parquet logical type; pretending we have a richer null-aware model than Parquet does invites bugs at boundaries.
- **Few real users.** Production geospatial pipelines almost always know their CRS; "unknown CRS" is mostly an artifact of bad ingest tooling, and those tools can target 1.x.

### For keeping `null`

- **1.x compatibility.** Users migrating 1.x → 2.0 may have files with `crs: null` and expect the same semantics.
- **Real data has unknown CRS.** Legacy shapefiles without `.prj`, scraped data, OCR-derived geometry, intermediate processing steps — all real, all have undefined CRS.
- **Loss of expressiveness.** "Unknown" and "OGC:CRS84 (default)" are genuinely different facts about the data. Collapsing them silently can lead to wildly wrong analyses (treating projected coordinates as lon/lat).
- **The awkwardness is small.** One paragraph in the spec, one branch in writers. Not nothing, but not a lot either.

## Open questions

- Should 2.0 readers encountering a 1.x-style `null` reject the file, warn, or treat it as OGC:CRS84? (Affects migration paths.)
- Is there appetite to coordinate with the Parquet core spec to add a native "unknown CRS" sentinel, which would let GeoParquet keep `null` cleanly?
- If we keep `null` in 2.0, should we at least RECOMMEND that writers prefer either an explicit `<authority>:<code>` or omission, and use `null` only when truly necessary?

## Changes in this PR (relative to #284)

- `format-specs/schema.json`: remove the `{"type": "null"}` clause from the `crs` `oneOf`.
  - Column-metadata `crs` type: `object|string|null` → `object|string`.
  - Replace the `null` paragraph in §`crs` with a short paragraph explaining that 2.0 has no "unknown CRS" form and directing such writers to GeoParquet 1.x.
  - Remove the `null` row from the mirroring table.
- `scripts/test_json_schema.py`: flip `crs_null` from a valid to an invalid case.

```
57 passed in 0.5s  (test_json_schema.py)
72 passed in 2.1s  (full scripts/ test suite)
```

### DIFF
--- a/format-specs/geoparquet.md
+++ b/format-specs/geoparquet.md
@@ -49,7 +49,7 @@ Each geometry column in the dataset MUST be included in the `columns` field abov
 | -------------- | ------------ | ----------- |
 | encoding       | string       | **REQUIRED.** Name of the geometry encoding format. Only `"WKB"` is supported. |
 | geometry_types | \[string]    | **REQUIRED.** The geometry types of all geometries, or an empty array if they are not known. |
-| crs            | object\|string\|null | [PROJJSON](https://proj.org/specifications/projjson.html) object, or an `<authority>:<code>` string (e.g. `"EPSG:4326"`), representing the Coordinate Reference System (CRS) of the geometry. The value MUST equal the Parquet logical-type `crs` property in the same form. If the field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means the data in this column must be stored in longitude, latitude based on the WGS84 datum. |
+| crs            | object\|string | [PROJJSON](https://proj.org/specifications/projjson.html) object, or an `<authority>:<code>` string (e.g. `"EPSG:4326"`), representing the Coordinate Reference System (CRS) of the geometry. The value MUST equal the Parquet logical-type `crs` property in the same form. If the field is not provided, the default CRS is [OGC:CRS84](https://www.opengis.net/def/crs/OGC/1.3/CRS84), which means the data in this column must be stored in longitude, latitude based on the WGS84 datum. |
 | orientation    | string       | Winding order of exterior ring of polygons. If present must be `"counterclockwise"`; interior rings are wound in opposite order. If absent, no assertions are made regarding the winding order. |
 | edges          | string       | Describes how to interpret the edges of the geometries. Must be one of `planar`, `spherical`, `vincenty`, `thomas`, `andoyer`, `karney`. The default value is `planar`.
 | bbox           | \[number]    | Bounding Box of the geometries in the file, formatted according to [RFC 7946, section 5](https://tools.ietf.org/html/rfc7946#section-5). |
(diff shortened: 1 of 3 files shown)


m-mohr:
I think I'd also remove it. Without a CRS the values are pretty meaningless unless you have some semantics defined outside of GeoParquet. So maybe just don't list the column in the GeoParquet metadata in this case?

jorisvandenbossche:
I would prefer to keep this option. There are real-world valid use cases for having geometry data without a CRS, that are not just legacy data with unknown crs (e.g. data measured relative to a local reference, non-EO images like microscopes, ..).

Those probably could all define a custom engineering CRS? But that might not always add actual value, and at that point cause unnecessary overhead / complexity for the user having to define this.

m-mohr:
@jorisvandenbossche Would such data write GeoParquet metadata or just use the native types and what they support?

cholmes:
We had a good discussion of this on the call on June 1st. We discovered that geopackage use SRID:0 to communicate that an srs is unknown and cartesian, and srid:-1 to communicate that an srs is unknown geographic coordinate system.

We decided to keep our 'null' semantics in geoparquet metadata for backwards compatibility, and specify that the source parquet should use srid:0 (which is actually mentioned in the parquet geospatial spec, though it doesn't say what to use it for).

cholmes:
closing in favor of #286, which came from discussions of the core group.

Issue #279, opened 2026-01-09
User-defined Parquet index for spatial data


paleolimbot:
This was brought up in https://github.com/opengeospatial/geoparquet/pull/278 but deserves its own issue!

> It is! https://datafusion.apache.org/blog/2025/07/14/user-defined-parquet-indexes/ . This and the column chunk approach would need some prototyping and low-level engine integration that I don't currently have the bandwidth to do, but if you're interested feel free to open a dedicated issue and we can discuss/iterate! (There are a number of files at https://geoarrow.org/data that are already available in native/wkb/new types if you are looking for more targets to benchmark).

I've reviewed this blog post (thanks a bunch for that link, I had no idea this was allowed) and I think I can allocate some of my time to build some prototypes using `arrow-rs/parquet` and likely `datafusion` as I'm already at least somewhat familiar with both packages. I don't want to derail and muddy up the discussion here more than necessary; what would be the preferred way to collaborate on these prototypes?

_Originally posted by @BlakeOrth in https://github.com/opengeospatial/geoparquet/issues/278#issuecomment-3730374177_

The gist of that blog post is that it's legal to use the bytes inbetween row groups to store whatever you want (the footer references byte ranges in the file and doesn't care about anything that isn't within those byte ranges as far as I know). This means we could pack an index into the file.

I think right before the footer makes sense (because then we could potentially write a utility that rewrites a file by removing the footer, writing the index, then rewriting the footer). I'm not sure how other indexes have been implemented in this way and there's probably more to draw from.

> I don't want to derail and muddy up the discussion here more than necessary; what would be the preferred way to collaborate on these prototypes?

Maybe a pull request into geoarrow-rs? I'm not sure DataFusion is required for the prototype but a pull request into apache/sedona-db or datafusion-contrib/geodatafusion would work if it is. Or your own repo (I'm just excited that anybody is working on this!)

BlakeOrth:
Thanks for opening this for additional discussion!

> I think right before the footer makes sense (because then we could potentially write a utility that rewrites a file by removing the footer, writing the index, then rewriting the footer). 

I like the thought of being able to take any old Parquet file with GEO-type columns and being able to effectively enhance query performance by appending an additional index of some sort. Although given the details in the blog post it almost seems like the "where" in the file is more or less inconsequential as long as readers agree on a mechanism to locate its byte offset.

[...]

PR #288, opened 2026-06-18
Update distribution best practices guide for GeoParquet 2.0

cholmes:
## Update the distribution best practices guide for GeoParquet 2.0

Reworks the GeoParquet distribution guidance for 2.0 and splits it into two pages: a focused **recommendations** guide and a companion **tools** guide.

### Recommendations — `format-specs/distributing-geoparquet.md`
- Lead with **GeoParquet 2.0**: native Parquet `GEOMETRY`/`GEOGRAPHY` types carry per–row-group geospatial statistics, giving efficient spatial access without the extra `bbox` column 1.1 required (smaller files, especially for points). The 1.1 `bbox` covering is kept as a documented fallback.
- Renamed "bbox covering" → **Efficient spatial access**; updated Spatial Ordering and Frontend sections accordingly.
- **Compression**: recommend zstd **level 15 or higher**, going as high as time allows (reconciles the old tl;dr-vs-prose mismatch).
- **Overture** noted as distributed in GeoParquet 1.1.
- **STAC** recommendation now points to geoparquet-io plus rustac/pystac.
- New **Further Discussion → Page-level spatial statistics** section: explains row-group vs page-level pruning and how page-level geometry stats could be added to Parquet, summarizing the benchmarks and approaches from #279.

### Tools — `format-specs/distributing-geoparquet-tools.md` (new)
Per-tool how-tos, each covering conversion, native/2.0 output, spatial partitioning, and STAC where supported. Verified against current releases:
- **GDAL/OGR** (3.13): `USE_PARQUET_GEO_TYPES` (`NO`/`YES`/`ONLY`) for native types — `ONLY` recommended as the interim path (GDAL still writes 1.1 metadata; `ONLY` output is readable by 2.0 readers but not conformant 2.0). Spatial partitioning via `gdal vector partition`. No STAC output.
- **DuckDB** (1.5): `GEOMETRY` + GeoParquet I/O are core (CRS preserved on round-trip); `GEOPARQUET_VERSION 'V2'` writes conformant 2.0; `ST_*` (Hilbert ordering, reprojection) need `LOAD spatial`. Spatial partitioning example using the **a5** community extension + `COPY ... PARTITION_BY`. No STAC output.
- **geoparquet-io** (1.3): applies the best practices by default (zstd 15, Hilbert, bbox, 100k row groups); `--geoparquet-version 2.0` for 2.0. Spatial partitioning (`gpio partition`) and STAC (`gpio publish stac`) both documented.
- **Sedona**: KD-tree spatial partitioning example; other distribution settings still need a follow-up PR.

The two pages cross-link; all internal and cross-file anchors were verified.

> Note: this branch is based on `main` (still `2.0-dev`) and is independent of the separate version-bump branch. It can merge before or after that one.


jorisvandenbossche:
> Lead with **GeoParquet 2.0**

What is the status on support for this in general? You mention GDAL/DuckDB/geoparquet-io/Sedona as examples to generate the data, and so suppose those can also read such data (or at least with their latest versions). But outside of those, do we think the native types are supported widely enough to have that be the first recommendation to use? 
(I know GeoPandas is a bit behind ..)

And also if we recommend that for distributing, we should probably add a note on compatibility (eg recommend to still use GeoParquet 1.1 if you need to support a wider range of software/versions to read your data)

cholmes:
> What is the status on support for this in general? You mention GDAL/DuckDB/geoparquet-io/Sedona as examples to generate the data, and so suppose those can also read such data (or at least with their latest versions). But outside of those, do we think the native types are supported widely enough to have that be the first recommendation to use?

Yeah, they can all read the data. QGIS will read the data with gdal, and most tools use gdal under the hood. GDAL can read native parquet with version 3.12 I believe, so it's had some time to get out to other tools. Hyparquet and duckdb wasm are both read it, and those seem to be the most widely used tools in JS land. I obviously lean a bit towards making it the first recommendation, since it's what I wrote, but I'm also ok if we tone it down and stick with 1.1 as the recommendation here for a bit - until geopandas gets it :) And I'd guess Esri probably isn't there yet, so that could also be one we wait for. And also until we have more data sets to point out. Though I tried to say that 1.1 is fully acceptable for distributing.

>And also if we recommend that for distributing, we should probably add a note on compatibility (eg recommend to still use GeoParquet 1.1 if you need to support a wider range of software/versions to read your data)

Sounds good. Would be good to get the list of those who don't support it yet - both to inform users, and so we know who to try to get to upgrade.

format-specs/distributing-geoparquet-tools.md
```diff
@@ -0,0 +1,311 @@
...
+
+```
+ogr2ogr out.parquet in.geojson
+```
+
+Out of the box GDAL/OGR defaults to snappy compression, with max row group size of 65536.
+Version 3.9 and later will write out the `bbox` column by default, producing GeoParquet 1.1. And there is a built-in
+option to spatially order the data that works by creating a temporary GeoPackage file and
+using its r-tree spatial index. It defaults to false since it can be an intensive operation,
+and GDAL is usually translating from formats that already have spatial indexes.
```
paleolimbot:
```suggestion
Version 3.9 and later will write out the `bbox` column by default, producing GeoParquet 1.1. There is a built-in
option (`SORT_BY_BBOX=YES`) to spatially order the data that works by creating a temporary GeoPackage file and
using its r-tree spatial index. It defaults to false since it can be an intensive operation,
and GDAL is usually translating from formats that already have spatial indexes.
```

format-specs/distributing-geoparquet-tools.md
```diff
@@ -0,0 +1,311 @@
...
+
+#### GDAL/OGR with recommended settings
+
+These examples are done with the `ogr2ogr command-line tool, but the layer creation options
+will be the same calling from C or Python.
+
+You can easily control the compression and the max row group size, and the following command is sufficient
```
paleolimbot:
```suggestion
You can control the compression and the max row group size, and the following command is sufficient
```

format-specs/distributing-geoparquet-tools.md
paleolimbot: (Or just remove this part since you mention it later)

format-specs/distributing-geoparquet-tools.md
```diff
@@ -0,0 +1,311 @@
...
+if your source data is already spatially ordered in a file format with a spatial index (like FlatGeobuf or GeoPackage):
+```
+ogr2ogr out.parquet -lco "COMPRESSION=ZSTD" -lco "MAX_ROW_GROUP_SIZE=100000" in.fgb
+```
+
+GDAL 3.12 and above introduces `COMPRESSION_LEVEL` as a [Parquet layer creation option](https://gdal.org/en/latest/drivers/vector/parquet.html#layer-creation-options). So if you're working with that then you should definitely use it (along with the new
+[gdal CLI](https://gdal.org/en/latest/programs/index.html#general), which is used here.
```
paleolimbot:
```suggestion
GDAL 3.12 and above introduces `COMPRESSION_LEVEL` as a [Parquet layer creation option](https://gdal.org/en/latest/drivers/vector/parquet.html#layer-creation-options) and a new
[CLI](https://gdal.org/en/latest/programs/index.html#general), which is used here.
```

format-specs/distributing-geoparquet-tools.md
```diff
@@ -0,0 +1,311 @@
...
+```
+ogr2ogr out.parquet -lco SORT_BY_BBOX=YES -lco "COMPRESSION=ZSTD" in.geojson
+```
+
+This operation writes the data to a GeoPackage as an interim step, so it can take additional storage and computation, especially
+with large files, so it's not enabled by default.
+
+#### Writing native geometry types
+
+As of this writing GDAL does not yet write GeoParquet 2.0 metadata — by default it produces GeoParquet 1.1 with a `bbox`
```
paleolimbot:
```suggestion
As of this writing GDAL does not yet write GeoParquet 2.0 metadata: by default it produces GeoParquet 1.1 with a `bbox`
```

format-specs/distributing-geoparquet-tools.md
```diff
@@ -0,0 +1,311 @@
...
+
+Interestingly you can also set the row group size in bytes, which would likely be a better way to handle geospatial data since the
+row size can vary so much.
+
+```
+COPY (SELECT * FROM geo_table) TO 'out.parquet' (FORMAT 'parquet', GEOPARQUET_VERSION 'V2', COMPRESSION 'zstd', ROW_GROUP_SIZE_BYTES '128mb');
+
+```
+
+But you can only use that when [`SET preserve_insertion_order = false;`](https://duckdb.org/docs/stable/guides/performance/how_to_tune_workloads#the-preserve_insertion_order-option) is enabled, which can help when working with large files, but it's not
```
paleolimbot:
```suggestion
The `ROW_GROUP_SIZE_BYTES` option may only be used when [`SET preserve_insertion_order = false;`](https://duckdb.org/docs/stable/guides/performance/how_to_tune_workloads#the-preserve_insertion_order-option) is enabled, which can help when working with large files, but it's not
```

format-specs/distributing-geoparquet-tools.md
```diff
@@ -0,0 +1,311 @@
...
+clear if it can mess up spatial ordering.
```
paleolimbot:
```suggestion
clear if it can preserve spatial ordering.
```

paleolimbot (commented):
Thanks! All my comments are optional...mostly just highlighting sentences with informal language.

I'll PR to add SedonaDB here following the formula you have for the others (on top of this one or as a new PR if this merges first).

paleolimbot (approved):
Thank you!

PR #293, opened 2026-07-17
Add zstd compression to example storing GeoParquet with sedona

JakobMiksch:
The text in the changed README suggests to use zstd compression. The example at the bottom, where Sedona stores a GeoParquet, the zstd is not explicitly set. I learned in this [thread](https://github.com/apache/sedona/discussions/3109), that zstd compression can be set using `option`. I tested it and it works! This PR adapts the example accordingly.


jiayuasu (approved):

jiayuasu (approved):
Thank you very much!

Issue #291, opened 2026-06-30
Support for Web Clients?


mmgeorge:
Hi! We've been working on adding display optimization that builds on top of GeoParquet, with the goal of enabling web clients to read and visualize GeoParquet files directly by streaming via range requests. I was wondering if there was any interest in standardizing any of this into the GeoParquet spec directly.

The approach we're using I think is actually pretty straightforward:  
- For points we simply cluster the dataset using Z-order.
- For non-points, we cluster using an XZ-order code derived from feature extents, using the approach outlined in https://doi.org/10.1007/3-540-48482-5_7. Then we generate several multiscale columns that include quantized, delta-encoded geometries to handle different LODs. We write these multiscale columns, and the XZ-key, as an additional structured column.

Because the XZ-code is an actual column, clients can use statistics to query the XZ-code ranges relevant for a given spatial query (like an on-screen display tile). Provided the file was generated with a Page Index (we currently require this), we are seeing pretty decent performance.

This is the full Overture building footprints dataset, (2.7 billion features spread across 16 files, each about ~25 GB):

https://github.com/user-attachments/assets/e76cd2a5-6d75-4992-a5b4-1ec8c930cf2f

[Overture Building Footprints](https://codepen.io/matt9222/pen/dPNVXZv)

And here's a single file containing 2010 US Census Blocks (11 million features in a single 10 GB file):

https://github.com/user-attachments/assets/3c18dde1-de77-482b-9901-31c172616bbb

[Census Blocks](https://codepen.io/matt9222/pen/pvRWbpB)

Right now we are storing this provisional "Spatial Optimization" metadata as a key-value pair in addition to the GeoParquet metadata (which covers the full resolution geometry). 

I noticed there's some other exploratory work going on in this area (https://github.com/opengeospatial/geoparquet/issues/279). Where this approach potentially differs from that is that while clustering spatially in this way does lead to better spatial query performance by colocating nearby features, it does fundamentally come at the expense of other variables you might want to cluster on. Storing additional LOD columns also means increasing the size of your dataset, so there are some real tradeoffs here. That said, if your data is spatial and you are using spatial filtering or performing spatial analysis, I imagine standard query engine predicate skipping should be able to benefit from this as well, provided we also write per-feature bounds that they could take advantage of (right now we are not doing this).

[...]

PR #294, opened 2026-07-18
Add support for display optimization

mmgeorge:
Opening this to give a more concrete spec proposal to go with #291 and see if there is interest! Wasn't sure if I should bump to 2.1 as part of this proposal or not, given that 2.0 is not yet released. 

Not sure if "display optimization" is the right framing, as any query engine should be able to take advantage of the spatial ordering. It could make senses to remove `display` and move that up one level, with an optional `ordering` and `lods` at the top level, and remove `geomery_column` in favor of just the `primary_column`. 

If there is interest in this, I can adapt the reference writer that I'm planning to release as part of the SOP work to output this as well, so that there's some implementation that can be referenced. I opted not to put in algorithm blocks, but open to adding those as well. 

I'm also experimenting with a quantized integer version of the (now removed) native encoding of 1.1, as that would hit DELTA_BINARY_PACKED which is similar to the PBF varint encoding.


paleolimbot (commented):
Thank you for opening!

In general I think this is a fantastic idea...using varint encodings for geometry has come up independently ( https://rusty.today/blog/geosilo-vs-geoarrow-byte-stream-split-geometry-compression/ ), and XZ encoding was discussed as a potential sorting when we were negotiating the Parquet and Iceberg geometry/geography specs. Being able to display Parquet files effectively has been a challenge and this is the fastest I've ever seen the rendering happen...it's impressive!

At a high level I am wondering if we can encode some of these concepts into existing Parquet concepts so that we can find a path for them in the Parquet spec. While we can definitely keep iterating on the metadata specification, I think we are in general trying to move towards a place where readers no longer need it.

Off the top of my head, I wonder if something like this would work:

- Add the min/max of the `xz` value to `GeoStatistics` (it would be in relation to the bounds already listed in the GeoStatistics for x and y...I don't think there's an easy way to store a file-level bounds from which all XZ values would be derived, but maybe we'd need that for this to be effective)
- Add a page-level index that contains GeoStatistics values for pages. We want this anyway because right now people are making row groups tiny to compensate which slows down bulk reads.
- Add an "encoding" for geometry and geography values with the pbf encoding you described or similar. It's common for query engines to operate on encoded data as an optimization, even though for most pathways the data would be decoded into WKB and that's what query engines would see.

All these things can be pursued independently of eachother (including adding this as part of the metadata spec!)

I'd love to hear other thoughts on this as well!

mmgeorge:
Interesting! I'll have to take a closer look at that format. 🤔 If there's a path to get support for the different in Parquet directly, definitely agree with going that route! 

For XZ, I don't know that it necessarily even has to be a spatial-only concept. It's essentially just a way to cluster ranges, so I think it could even be applicable for non-spatial datasets with just something like a temporal interval (start/end time). There already is a somewhat similar concept of [sorting_columns](https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L1044) in the spec, so maybe all we need is metadata about how the file was clustered (method + fields), and then those clusterKeys could be stored in the page metadata like you are suggesting. This might also help the table formats that support clustering. 

I think I also need to try using covering page stats (instead of XZ ranges) & see if there is a tradeoff there (might actually perform better). I do currently use a per-feature XZ code though to know whether I should skip features within page. 

The main sticking point I think might be the LOD columns which sort of extends beyond just encoding, since we are actually generalizing and writing additional lower-resolution versions of the geometry for display. I'm not sure that a query engine would ever want to look at these, just for visualization I think. That said, stuff like the resolution and scale that levels are valid for, along with any transformations, could theoretically be moved to column metadata (like the spatial reference info for the geometry/geography types).

cholmes:
Thanks for the contribution! +1 to try to shape things for ease of integration in the Parquet spec, and also for pursuing the pieces independently. 

As for the spec itself, I lean towards putting this in its own spec, at least to start, versus bumping the version to 2.1. It can be similar to the raster spec, where it incubates, and then can be an 'extension'. 

Are there any sample files to experiment with? Would be good to have someI recently put up https://github.com/geoparquet/geoparquet-testing and we could use that, perhaps in an 'incubating' folder to indicate that it's not yet part of the spec. 

I think the main thing we'll need to actually merge this in is to have at least 3 different implementations. It does seem like there's a couple other related sets of work, with https://github.com/Kanahiro/cloud-optimized-geoparquet and https://github.com/yharby/geoparquet-overviews - we should evaluate the different approaches and see what could work best for all use cases.

mmgeorge:
> I wonder if XZ ranges per page of what you actually care about (i.e., the protobuf-encoded geometries in this case) would work better also because the page boundaries would align perfectly. I am not sure exactly of the internals of the writer, but I gather that pages for different columns don't necessarily have the same boundaries (so your xz-derived row ranges might still fetch too many pages in a lot of cases).

Right exactly, the writer I have doesn't do anything special -- it's just using DataFusion. The code adds an XZ column computed from feature extents (wkb geometry), sorts by this column, and then add the "geodisplay" column with the XZ code and multiscale columns. That outputs a Parquet file with the defaults of whatever the writing engine requires (GeoAnalytics Engine is the production version of this based on Spark). We don't have control over the row groups / page sizes that different enviornments use (esp. managed cloud), so we wanted to avoid anything that would require modifying those. A very early prototype I had did use small row groups, but that blows up file size. 

Hopefully getting this writer pushed to the Esri org soon! (My fault, lost access to my 2FA for my work github).

The reader we have integrated into our engine is based on https://github.com/jorgecarleitao/parquet2 since the custom async thrift reader it uses is web friendly and the Wasm size is pretty small. Currently evaluating whether we can move to https://github.com/apache/arrow-rs/tree/main/parquet (without the arrow feature which pulls in a lot of code). As we pan around the map, tile extent queries are passed down & then converted to XZ code ranges. Then we prune by file stats computed from row groups (for Overture buildings we can have several files), then row groups. If row group passes -> we lazy load the page index for the XZ column. Then we find which pages passes. 

Like you mentioned, the current problem is that pages for XZ values do not align at all with the actual quantized geometry. Making up numbers here, but a geometry page might have only 10 features, whereas an XZ page might have the codes for 10,000. That means that we have to go one step further and grab actual page values when building our selection bitset, rather than just selecting all rows in the page. Aligning XZ codes with the requested LOD geometry would definitely help here. We may still want to pull in the actual values though? The reason being that the XZ/the spatial filter is just one predicate -- we also support setting a definintionExpression (where clause) on the layer itself, which can also affect the selection bitset. So we repeat the process for the where clause, streaming in just the columns that we need for it (unfortunately again the row ranges of each page will not match at all), and only after that do we actually touch geometries. 

I have some dev code can also read file level stats that come from an external JSON file, but I think we'd prefer to just support Iceberg instead. 

> As for the spec itself, I lean towards putting this in its own spec, at least to start, versus bumping the version to 2.1. It can be similar to the raster spec, where it incubates, and then can be an 'extension'.

> Are there any sample files to experiment with? Would be good to have someI recently put up https://github.com/geoparquet/geoparquet-testing and we could use that, perhaps in an 'incubating' folder to indicate that it's not yet part of the spec.

Yes I think that makes a lot of sense! Will work on the changes. I have some files but they are large -- let me generate some smaller test data. I need to update my writer to be able to output in the format outline above (the current SOP spec has metadata in more of an Esri-way, with things like WKID / WKT / hasZ / hasM, tried to bring that in line with the main GeoParquet metadata).

> I think the main thing we'll need to actually merge this in is to have at least 3 different implementations. It does seem like there's a couple other related sets of work, with https://github.com/Kanahiro/cloud-optimized-geoparquet and https://github.com/yharby/geoparquet-overviews - we should evaluate the different approaches and see what could work best for all use cases.

There are both very interesting approaches. In some ways they actually turn Parquet into a display format 🤔. These are more similar to our "feature reduction" or "feature thinning" concepts -- a given extent does not include all the features, instead it thins it out to improve performance. I did a thinning POC (for traditional Esri Feature Service backed Feature Layer, not Parquet Feature Layer), but one of the problems we quickly ran into is that for operational data, users really want "all the data" for a given extent. This is especially true for dashboard-style apps. The reason being is that they want to run some kind of analysis, e.g.: 

https://github.com/user-attachments/assets/5632148d-9990-41bf-ae40-82d665d7c3c9

https://developers.arcgis.com/javascript/latest/sample-code/featurelayerview-query-geometry/?view=preview-only

or the may want to "cluster" (visually) the data: 

Screenshot 2026-07-19 at 12 53 42 PM

https://developers.arcgis.com/javascript/latest/sample-code/featurereduction-cluster-filter-slider/?view=preview-only

Also when you symbolize polygons with markers, you get behavior that looks buggy, with markers disappearing as you zoom out.

But at the same time, some datasets are sometimes just "too large" and having some way to thin out the data is another interesting option to have at your disposal... and something that maybe you just need to require to be on at a certain scale. Need to try these approaches out!

mmgeorge:
Closing in favor of https://github.com/opengeospatial/geoparquet/pull/295

PR #295, opened 2026-07-20
Add display extension proposals

mmgeorge:
https://github.com/opengeospatial/geoparquet/issues/291 

This PR splits up the changes from https://github.com/opengeospatial/geoparquet/pull/294 into two separate smaller extensions. If display is not a goal, implementors can just sort spatially. It also should also hopefully make transitioning to possible future parquet-native functionality easier. E.g., if spatial sorting gets pushed down to parquet directly, then we can transition out that extension. It also allows us more granularity in potentially exploring different and combining different visualization approaches, such as feature thinning.


mmgeorge:
The provided parquet files are 1.1, upgrading to 2.0 now that the rc has been cut.

mmgeorge:
Hmm good point, scale is more 2D centric.

mmgeorge:
Logical pixels per map unit (resolution) is actually currently the transform x & y right now in the data I've generated. But if you wanted to quantize with greater fidelity (to avoid quantization artifacts when zooming in), they don't have to be equal. 

If you are rendering with tilt, you probably want to look at the projected size in pixels of a tile of features you are rendering (or more granularly you could look at the projected bbox of each feature), then compare that to the map size of the tile, and then select the closest / more detailed LOD whose resolution matches the projected resolution of the tile / bbox. Let me flesh this out, I think resolution might be the better thing to write, esp. since scale is based on the magical physical size of a pixel that you end up coming up with ([ArcGIS uses 96 DPI](https://developers.arcgis.com/documentation/mapping-and-location-services/reference/zoom-levels-and-scale/), [OGC uses 0.28](https://docs.ogc.org/is/17-083r4/17-083r4.html))

mmgeorge:
I have not forgotten about this! Been working on updating the viewer code to work with this proposal, support GeoParquet 2.0, and switch to using resolution instead of scale. Should be able to get back to this PR next week.

mmgeorge:
> We were wondering about the placement of the metadata. Should the ordering and lods metadata be at the top level and reference geometry_columns? Or should the existing geometry_columns inline the lod and ordering metadata?

Good point, that makes geometry_column unnecessary: 

```json
{
  "primary_column": "geometry",
  "columns": {
    "geometry": {
      "encoding": "WKB",
      "geometry_types": ["Polygon", "MultiPolygon"],
      "ordering": {
        "type": "xz",
        "extent": [-180, -90, 180, 90],
        "max_level": 20
      },
      "lod": {
        "encoding": "pbf",
        "orientation": "clockwise",
        "levels": [
          {
            "column": ["geolod", "level_1"],
            "resolution": 0.01,
            "transform": {
              "scale": [0.01, 0.01, 1, 1],
              "translate": [0, 0, 0, 0]
            }
          }
        ]
      }
    }
  }
}
```

For LODs I think it makes sense -- that is scoped to the geometry, and you could potentially have multiple sets of geometries with their own LODs (if you wanted to include multiple projections for instance).

Ordering might still make sense at the top level since you can only have one physical ordering, and we also probably want to support for including other things in the cluster key? But maybe we should rename geometry_column -> columns: string[] and specify that columns for now must be a single geometry (for now)? 

```json
"ordering": {
    "type": "xz",
    "columns": [ "geometry" ],
    "extent": [-180, -90, 180, 90],
    "max_level": 20
  }
```
 
In the future: 

```json
"ordering": {
    "type": "xz",
    "columns": [ "geometry", "instant" ],
    "extent": [-180, -90, 180, 90],
    "max_level": 20
  }
```

Though now that I'm thinking about it, you might also want something like: 

```json
"ordering": {
    "type": "xz",
    "columns": [ "geometry", { "range_start":  "start_time",  "range_end": "end_time" } ],
    "extent": [-180, -90, 180, 90],
    "max_level": 20
  }
```

^ not a real proposal, suggests that maybe this needs to be framed as a set of operations? Or I suppose we could have a bunch of well defined types like "xz", "xzt", "xzi" (internal), etc.

If querying on the client can work without actually needing to be aware of the actual clusterKey, then we might not need to encode this information at all. 

------

> Up to this point the GeoParquet specification hasn't really defined its own geometry representation...we took existing specifications (GeoArrow, WKB, Parquet, PROJJSON) and connected them, but all of those are maintained elsewhere and we just reference them by name. I also get that the protobuf format you've invented here is a vast improvement and is key to reducing the IO requirement from the client. I'm not sure exactly what the answer is here but I am sure we can find a way to make this work.

The format is actually just a subset of https://github.com/Esri/arcgis-pbf, which I could improve the doc for (and we could link out to that if it's preferred), esp. on the geometry message part. 

But I've also been trying to experiment with a few other options and to get some actual numbers that we can compare to. I'm not sure if the ship has completely sailed yet on the native encoding, but using that format but with i64 + delta_binary_packed appears to be very competitive with pbf at least with snappy. A bit worse with gzip / zstd, but there actually the standard native encoding actually performs the same as the integer one. 

Then with either of these native encodings we could just rely on page stats. And on the client side a columnar Z/M that you can avoid downloading does have it's advantages. 

# Dataset: [us_schools](https://www.census.gov/geographies/mapping-files/time-series/geo/tiger-geopackage-file.html)

## Compressed Size, File

| Format | Encoding | Compression | Size (MiB) | Vs wkb % |
| --- | --- | --- | --- | --- |
| wkb | delta_byte_array | snappy | 510.94 | +0% |
| wkb-quantized | delta_byte_array | snappy | 482.12 | -6% |
| native | byte_stream_split | snappy | 398.79 | -22% |
| native-quantized-float | byte_stream_split | snappy | 348.09 | -32% |
| wkb | delta_byte_array | zstd | 336.35 | +0% |
| wkb | delta_byte_array | gzip | 335.91 | +0% |
| native | byte_stream_split | zstd | 317.49 | -6% |
| native-quantized | delta_binary_packed | snappy | 302.30 | -41% |
| wkb-quantized | delta_byte_array | gzip | 298.50 | -11% |
| pbf | delta_byte_array | snappy | 295.30 | -42% |
| wkb-quantized | delta_byte_array | zstd | 288.71 | -14% |
| native | byte_stream_split | gzip | 283.92 | -15% |
| native-quantized-float | byte_stream_split | zstd | 266.26 | -21% |
| native-quantized | delta_binary_packed | zstd | 260.99 | -22% |
| native-quantized | delta_binary_packed | gzip | 237.30 | -29% |
| native-quantized-float | byte_stream_split | gzip | 237.14 | -29% |
| pbf | delta_byte_array | zstd | 234.92 | -30% |
| pbf | delta_byte_array | gzip | 215.28 | -36% |

## Compressed Size by LOD, snappy

| Format | Encoding | Full (MiB) | LOD total (MiB) | level_2 | level_4 | level_6 | level_8 | level_10 | level_12 | level_14 | level_16 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| wkb | delta_byte_array | 218.48 | 291.85 (+0%) | 0.38 (+0%) | 1.12 (+0%) | 3.43 (+0%) | 9.11 (+0%) | 23.87 (+0%) | 52.50 (+0%) | 88.08 (+0%) | 113.23 (+0%) |
| wkb-quantized | delta_byte_array | 218.48 | 263.03 (-10%) | 0.36 (-6%) | 1.07 (-5%) | 3.29 (-4%) | 9.04 (-1%) | 21.86 (-8%) | 45.55 (-13%) | 75.36 (-14%) | 106.39 (-6%) |
| native | byte_stream_split | 218.48 | 179.67 (-38%) | 0.29 (-22%) | 0.91 (-19%) | 2.66 (-22%) | 6.35 (-30%) | 16.34 (-32%) | 34.11 (-35%) | 54.28 (-38%) | 64.64 (-43%) |
| native-quantized-float | byte_stream_split | 218.48 | 128.97 (-56%) | 0.22 (-42%) | 0.64 (-43%) | 1.88 (-45%) | 5.01 (-45%) | 10.97 (-54%) | 22.31 (-57%) | 36.13 (-59%) | 51.72 (-54%) |
| native-quantized | delta_binary_packed | 218.48 | 83.20 (-71%) | 0.14 (-63%) | 0.45 (-60%) | 1.32 (-62%) | 3.27 (-64%) | 7.16 (-70%) | 13.99 (-73%) | 23.43 (-73%) | 33.41 (-70%) |
| pbf | delta_byte_array | 218.48 | 76.21 (-74%) | 0.14 (-63%) | 0.35 (-69%) | 0.94 (-73%) | 2.39 (-74%) | 5.75 (-76%) | 12.65 (-76%) | 22.80 (-74%) | 31.12 (-73%) |

## Compressed Size by LOD, gzip

| Format | Encoding | Full (MiB) | LOD total (MiB) | level_2 | level_4 | level_6 | level_8 | level_10 | level_12 | level_14 | level_16 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| wkb | delta_byte_array | 157.94 | 177.52 (+0%) | 0.20 (+0%) | 0.62 (+0%) | 1.94 (+0%) | 5.12 (+0%) | 14.12 (+0%) | 31.84 (+0%) | 54.22 (+0%) | 69.40 (+0%) |
| wkb-quantized | delta_byte_array | 157.94 | 140.11 (-21%) | 0.18 (-7%) | 0.54 (-13%) | 1.72 (-11%) | 4.68 (-9%) | 11.48 (-19%) | 24.25 (-24%) | 40.10 (-26%) | 57.08 (-18%) |
| native | byte_stream_split | 157.94 | 125.50 (-29%) | 0.17 (-11%) | 0.54 (-13%) | 1.53 (-21%) | 3.58 (-30%) | 9.75 (-31%) | 22.05 (-31%) | 37.38 (-31%) | 50.44 (-27%) |
| native-quantized | delta_binary_packed | 157.94 | 78.90 (-56%) | 0.13 (-33%) | 0.39 (-37%) | 1.12 (-42%) | 2.80 (-45%) | 6.32 (-55%) | 12.96 (-59%) | 22.46 (-59%) | 32.69 (-53%) |
| native-quantized-float | byte_stream_split | 157.94 | 78.72 (-56%) | 0.12 (-38%) | 0.34 (-45%) | 0.97 (-50%) | 2.51 (-51%) | 5.60 (-60%) | 11.90 (-63%) | 22.85 (-58%) | 34.38 (-50%) |
| pbf | delta_byte_array | 157.94 | 56.89 (-68%) | 0.09 (-53%) | 0.23 (-63%) | 0.61 (-68%) | 1.56 (-69%) | 3.82 (-73%) | 8.65 (-73%) | 16.38 (-70%) | 25.50 (-63%) |

## Compressed Size by LOD, zstd

| Format | Encoding | Full (MiB) | LOD total (MiB) | level_2 | level_4 | level_6 | level_8 | level_10 | level_12 | level_14 | level_16 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| wkb | delta_byte_array | 177.50 | 158.39 (+0%) | 0.28 (+0%) | 0.78 (+0%) | 2.40 (+0%) | 5.69 (+0%) | 14.16 (+0%) | 28.94 (+0%) | 46.75 (+0%) | 59.27 (+0%) |
| native | byte_stream_split | 177.50 | 139.50 (-12%) | 0.21 (-26%) | 0.66 (-16%) | 2.08 (-13%) | 5.01 (-12%) | 12.20 (-14%) | 24.66 (-15%) | 40.53 (-13%) | 54.09 (-9%) |
| wkb-quantized | delta_byte_array | 177.50 | 110.75 (-30%) | 0.22 (-21%) | 0.57 (-28%) | 1.69 (-30%) | 4.34 (-24%) | 9.09 (-36%) | 17.60 (-39%) | 30.10 (-36%) | 47.05 (-21%) |
| native-quantized-float | byte_stream_split | 177.50 | 88.27 (-44%) | 0.13 (-53%) | 0.42 (-47%) | 1.31 (-45%) | 3.15 (-45%) | 6.46 (-54%) | 13.96 (-52%) | 25.61 (-45%) | 37.18 (-37%) |
| native-quantized | delta_binary_packed | 177.50 | 83.02 (-48%) | 0.14 (-52%) | 0.43 (-45%) | 1.30 (-46%) | 3.23 (-43%) | 7.07 (-50%) | 13.97 (-52%) | 23.43 (-50%) | 33.41 (-44%) |
| pbf | delta_byte_array | 177.50 | 56.96 (-64%) | 0.11 (-63%) | 0.25 (-68%) | 0.65 (-73%) | 1.60 (-72%) | 3.80 (-73%) | 8.54 (-70%) | 16.34 (-65%) | 25.63 (-57%) |

**native**: after quantization / simplification, apply the transform to get back to world space floats (no transform need to decode on client)
**native-quantized**: basically the same as pbf, just using native encoding format with i64 (transform required by client to decode)
**native-quantized-float**: still quantize / simplify, but write output with f64 instead of i64 (transform required by client to decode)
**wkb-quantized**: still quantize / simplify, but write output with f64 instead of i64 (transform required by client to decode)


---

Still need to try a few more datasets though.

mmgeorge:
We also could treat the multiscale columns as just regular "geometry" columns: 

```json
{
  "primary_column": "geometry",
  "columns": {
    "geometry": {
      "encoding": "WKB",
      "geometry_types": ["Polygon", "MultiPolygon"],
      "lods": [
        {
          "column": ["geolod", "level_1"],
          "resolution": 0.01,
        }
      ]
    }, 
    "geolod.level_1": { <- a bit ugly
        "encoding": "pbf",
        "encoding_transform": {
          "scale": ..., 
          "translate": ...,
        }
        "geometry_types": ["Polygon", "MultiPolygon"],
     }
  }
}
```

then encoding could be extended to support pbf, and we could introduce some kind of `encoding_transform` property

jorisvandenbossche:
My preference would be to treat this as several separate enhancements (new encoding, LOD columns, ordering info), although I don't know how feasible and desirable this is in practice (I can imagine that for your use case, for an ideal solution, you need the combination of features).

If we add the LoD concept, with an `encoding` for that simplified geometry column, in theory one could also use existing WKB encoding for that? (it would feel strange to disallow that, although I don't know how useful it would be in practice) 
And if we add a new "PBF"-like encoding for the LoD use case, assuming this can give a significant file size improvement, this would be interesting as a general encoding for the main geometry as well?

But in any case, @mmgeorge thanks for exploring this!

---

The reason that you are using an explicit "geokey" column included in the file, is that related to the fact that when you are viewing the data at a certain LoD, the simplified geometry column for that level does not have the geospatial statistics as the Geometry logical type has, and so therefore you are using a different solution for efficient spatial filtering?

mmgeorge:
I think that direction makes sense ... basically don't treat the LOD geometry columns as special, instead just use the existing geometry column definition & extend to support `encoding: <new format>`. And then `lods` maybe moves back to the top level and points to each column? Or we could avoid any top level "lods" object, and each geometry column could have an optional `resolution` parameter which clients could use to understand when a geometry should be used? The maybe potentially strange thing is that this could allow clients to say write one LOD with WKB and another with PBF, but implementation-wise I don't think there's any real challenge with supporting that. And it could be that at certain LODs we might find some encodings preferable to others. So there could actually be a reason to want to allow them to be varied. 

Covering could be a better way to frame this... it is essentially just a different encoding for the bounding box. 

Let me try spliting this up into a few different PRs. 

-----

For the geokey, I implemented the logic very close to [XZ-Ordering: A Space-Filling Curve for Objects with Spatial Extension](https://www.semanticscholar.org/paper/XZ-Ordering%3A-A-Space-Filling-Curve-for-Objects-with-B%C3%B6hm-Klump/9dd2076f0beb188738f031583d35a6f3b15f7437). In my current implementation, I go tile extent -> xz code range covering -> rows -> grab LOD pages with rows.

If PBF had page-level geospatial statistics, I could use that instead (GEOMETRY / GEOGRAPHY currently only have it at the row group level right now I believe, which isn't sufficient either). 

-----

I got a quantized-native version working (this only uses page stats and avoids the xz code all-together), but getting some mediocre results. Need to dobule check a few things & will post a bit more on this tomorrow. Smaller than WKB but can be 2x larger than PBF at some scales, so the savings from avoiding the geokey download don't add up.

mmgeorge:
Ok I was looking into the quantized native version a bit more because I'm a bit surprised it underperforms PBF by so much. Interestingly pre-compression, the encoding actually is more efficient up to LOD 12. Post-compression it performs much worse across the board (see the previous table above). That seems to be caused by the fact that for PBF, we store as <large_world_vertex, dx1, dx2, dx3>. That results in getting a bunch of really smalls deltas like +-0,1,2 which gzip then compresses to less than one byte. Whereas for DELTA_BINARY_PACKED, all redundancy bas basically been removed already and so it doesn't compress much more. 

Closing this, will reopen as separate PRs.

---

In the native decomposition column, below the first number is the base DELTA_BINARY_PACKED value for the X & Y column values, the second two values are the repetition / definition levels (which are unfortunately duplicated for each column, so this would be worse for Z & M. 

## Census

| LOD | Vertices/feature | PBF* | Native decomposition* | Native total* | Ratio |
|---:|---:|---:|---:|---:|---:|
| 0 | 1.0 | 6.99 | 1.70 + 0.38 + 0.38 | 2.45 | 0.35x |
| 2 | 1.2 | 15.92 | 3.31 + 1.05 + 1.05 | 5.42 | 0.34x |
| 4 | 1.7 | 24.56 | 5.61 + 1.77 + 1.77 | 9.14 | 0.37x |
| 6 | 3.2 | 26.65 | 8.07 + 1.95 + 1.95 | 11.98 | 0.45x |
| 8 | 7.5 | 24.15 | 10.56 + 1.70 + 1.70 | 13.96 | 0.58x |
| 10 | 17.6 | 19.79 | 12.87 + 1.21 + 1.21 | 15.29 | 0.77x |
| 12 | 33.6 | 18.13 | 15.05 + 0.84 + 0.84 | 16.73 | 0.92x |
| 14 | 47.8 | 17.91 | 17.76 + 0.67 + 0.67 | 19.10 | 1.07x |
| 16 | 55.6 | 18.92 | 21.18 + 0.61 + 0.61 | 22.40 | 1.18x |

## US Schools

| LOD | Vertices/feature | PBF* | Native decomposition* | Native total* | Ratio |
|---:|---:|---:|---:|---:|---:|
| 0 | 3.3 | 25.55 | 6.56 + 2.01 + 2.01 | 10.59 | 0.41x |
| 2 | 8.7 | 21.97 | 8.58 + 1.97 + 1.97 | 12.52 | 0.57x |
| 4 | 25.7 | 18.55 | 10.67 + 1.38 + 1.38 | 13.43 | 0.72x |
| 6 | 74.8 | 17.31 | 12.38 + 0.70 + 0.70 | 13.78 | 0.80x |
| 8 | 181.8 | 16.82 | 13.40 + 0.36 + 0.36 | 14.12 | 0.84x |
| 10 | 392.4 | 16.47 | 13.74 + 0.19 + 0.19 | 14.13 | 0.86x |
| 12 | 720.6 | 16.48 | 14.76 + 0.12 + 0.12 | 14.99 | 0.91x |
| 14 | 1051.6 | 17.25 | 17.02 + 0.09 + 0.09 | 17.20 | 1.00x |
| 16 | 1261.2 | 19.09 | 20.30 + 0.08 + 0.08 | 20.45 | 1.07x |

*In bits/vertex

jatorre:
Thanks for looking at this. I am overall happy with the idea of adding optional overviews in GeoParquet.

As drafted, `lod.encoding` MUST be `"pbf"`. That makes the overview columns opaque to every engine that is not this client. DuckDB, Sedona, GDAL, BigQuery, Snowflake, Databricks, none of them can do anything with a protobuf blob column. They will carry the bytes, replicate them across every copy of the dataset, and bill for them, and get nothing back. Or we are now forcing a lot of engines to understand PBFs, which I think is not realistic and not a good idea.

Plus I think simplified geometries can be useful beyond visualization, for example for coarse analysis. Approximate spatial joins, coarse aggregation, server-side rendering, quick overlap checks at continental scale, these all want a cheap low-vertex version of the geometry, and today people build them by hand every time. If the overview columns are ordinary geometry columns, every engine gets that for free and the feature pays for itself outside the browser. If they are PBF, the browser client is the only consumer that benefits.

It is also worth noting the direction 2.0 just took. 1.1 had several encodings and 2.0 deliberately narrowed to one: WKB in a `BYTE_ARRAY` with a `GEOMETRY`/`GEOGRAPHY` logical type. Adding a vendor protobuf as a required encoding reverses that consolidation. And I would like to see us aligned more and more with Iceberg, where WKB is the format for the geometry and geography types.

The other thing I would flag is that this PR encodes one particular overview strategy, and I do not think we should lock the spec to it yet. @paleolimbot mentioned above that @yharby has been working in this area, and it is worth looking at what he built (disclosure, Youssef is a colleague of mine): https://github.com/yharby/geoparquet-overviews. There is a draft convention, a reference converter and a live viewer. The approach is quite different: density-thinned bands laid out as a row-group prefix, with the simplified copy in a second ordinary geometry column listed in `geo.columns`, so the result is still a plain GeoParquet file that DuckDB reads unchanged. That feels more natural to me for the Parquet and Iceberg ecosystem, because nothing about the file is special outside the client that wants the fast path.

I do not want to hand-wave the cost. From your zstd numbers on US Schools, against the 177.50 MiB base file:

| overview encoding | LOD total | file growth |
| --- | --- | --- |
| pbf | 56.96 MiB | +32% |
| native-quantized | 83.02 MiB | +47% |
| quantized WKB | 110.75 MiB | +62% |
| WKB | 158.39 MiB | +89% |

So engine-readable overviews are genuinely more expensive, and I still think that is the right trade. Youssef's numbers are a useful comparison: a first paint that reads about 2 MB and decodes about 13 MB against a 620 MB exact geometry column, where nearly all of the win comes from thinning and from row-group and page pruning rather than from packing each vertex more tightly. The encoding looks like a second-order lever to me. It buys roughly 15% of base file size in your tables, while reading fewer features buys orders of magnitude. I would rather spend the interoperability budget on the first-order lever and keep the columns readable.

So overall I prefer a consistent single encoding to be honest. What should be optional here is whether a file carries overviews at all, not the format they are written in. If you write them, they should be WKB.

One thing this does not cost you: a WKB overview column is still a single `BYTE_ARRAY` column, so it keeps the page-locality property you correctly want, and it picks up Parquet geospatial statistics per row group as a side effect. It is only the split x/y layout that breaks locality, and that is not what I am suggesting.

mmgeorge:
Thanks for the comments @jatorre! I just want to be very clear -- I don't think we are at all talking about locking in a specific spec here. These PRs are just focused on adding proposal extensions. 

I think goals is also going to be a really important thing to think about here. At some LODs, WKB can be more than 3x larger than the equivalent PBF. Download size is hugely important when we're talking about web clients that are trying to directly visualize these features, and significantly less so for a back-end service reading directly from the same datacenter. So there's some real tradeoffs here that we need to consider. Definitely on our end, the focus is web clients. And if that's the goal, does it even make sense for something like this to be part of GeoParquet in the first place? That's one of the questions I think that we still need to work out. 

> As drafted, lod.encoding MUST be "pbf". That makes the overview columns opaque to every engine that is not this client

I think we are already in agreement that we should reuse geometry columns for this, in which case WKB would be a valid option here 👍 

> The approach is quite different: density-thinned bands laid out as a row-group prefix, with the simplified copy in a second ordinary geometry column listed in geo.columns, so the result is still a plain GeoParquet file that DuckDB reads unchanged. That feels more natural to me for the Parquet and Iceberg ecosystem, because nothing about the file is special outside the client that wants the fast path

Yes I think this is a really interesting approach! I actually think if we could combine spatial sorting, LODs and some kind of feature thinning (with clients able to determine how much thinning they want), then that would be really powerful. Have you guys opened any discussions with Iceberg to see what they think about it? The concern I would have is that changing the physical order of the file to support feature thinning sacrifices non-display query performance by spreading nearby features further apart. 

The approach we're using here is XZ-order, which I know has already come up in Iceberg discussions in the past. One of the things that I think is nice about XZ is that it really is a parquet-native way to go about this -- we're just applying a sort order. XZ already has some prior art for clustering geometries, and when paired with the full resolution WKB geometry column, any query engine should be able to take advantage of this even if they are unaware of the sort order or additional LOD columns. The LOD columns can then come into play for web clients who need to limit the amount of data they are streaming in.

I have some sample code that renders the SOP-spec version of this in Maplibre, hoping to get this up soon! And once we have the proposal extensions in place I can update it to support that as well.

PR #296, opened 2026-07-22
Add Geomermaids OSM daily exports to distribution examples

gsueur:
Adds the Geomermaids OSM daily exports to the list of GeoParquet data distribution examples in `distributing-geoparquet.md`.

These are daily OpenStreetMap exports published as GeoParquet v2.0, with Hilbert sorting and spatial partitioning, hosted at https://parquetry.geomermaids.com/. They serve as a real-world example of the distribution best practices described in this document.

### DIFF
--- a/format-specs/distributing-geoparquet.md
+++ b/format-specs/distributing-geoparquet.md
@@ -170,6 +170,7 @@ splits larger countries into smaller files, using S2 cells.
 Wherobots.
 - [Planet Ag Field Boundaries over EU](https://source.coop/repositories/planet/eu-field-boundaries/description) - needs to be
 spatially partitioned, row group size is 25,000.
+- [Geomermaids OSM daily exports](https://parquetry.geomermaids.com/) - GeoParquet v2.0 OSM exports with Hilbert sorting and partitioning.
 
 ## Tool Examples
 


cholmes:
Sorry for the slow response! Thanks for the contribution. I'm slammed right now, but will try to find some time in the next couple weeks to do at least a minimal bit of review on the OSM exports.

gsueur:
sure, no worries

cholmes:
Managed to find some time to check. It looks great! Though it does make me want to update the spec and best practices to articulate the benefits of adding bbox column to geoparquet 2.0. In gpio I give warnings now. So once we get it so that's in the spec I think it makes sense to merge in. I think there's rough consensus to add bbox covering back into the spec as optional, so I'll try to make that PR soon (or feel free to go for it with any of this). Thanks!

gsueur:
I don't know if it really needs to go down into the specs, but for remote usage on files > 500MB it definitely helps to hard prune on the bbox column. So it's not all use cases and probably better implementation of 2.0 on the reader side will help too. With my GeoPQ-Workbench I have no issue working with 2.0 remote files up to 3-4GB !

cholmes:
Yeah, I think it's worth having in the spec just so that if people want to do it then it's clearly an option for 2.0 and not some undocumented thing. I think we can have the best practices doc discuss what are the use cases where you'd actually want to use it. Your input here and anything more you want to put in is great - we can just say where it helps, and also the tradeoffs with it (ie increasing the overall size of your data).

gsueur:
Wrote up the bbox covering case in #297, with size and timing numbers from these files. Happy to keep this PR open until the spec side lands.

cholmes:
Hey, apologies for the super long delay - I finally got to some claude assisted review. Generally most things look good, ran analysis over much of it. 

Claude came up with two gaps, next two sections fully ai, but there's decent information. Though I disagree with claude on 'drop it', since I know you're aiming to show the usefulness of the bbox in 2.0 - but I don't really understand that it says no page index is written? I'd guess you left it off since you did this before we updated the spec to have the covering as optional.

### The miss: an undeclared bbox covering column

Every file carries a per-row `bbox` struct (`FLOAT` xmin/ymin/xmax/ymax), but it is **not declared in the `covering` field of the `geo` metadata**, so no reader will use it automatically. `gpio check` flags it independently:

```
⚠️  Bbox column 'bbox' found but not declared in 'covering' metadata
   As written, it costs file size and no reader can use it.
```

It also can't deliver the one benefit the guide still credits a covering with — page-level pruning — because **no page index is written at all** (no `ColumnIndex`/`OffsetIndex` on any column). That's a DuckDB writer default, not a choice here; a control re-write with DuckDB reproduces it.

The cost is real: the bbox column is **20.9%** of the places (point) file, **8.1%** of roads and **7.7%** of buildings — roughly **2 GB of the 27.5 GB snapshot, rebuilt nightly**. And the guide specifically calls out a bbox covering as "rarely worth adding" for point data, which `places` and `pois` are.

Good news is the values are correct — I verified against all 137,677 road geometries that the float bounds are rounded outward (0 violations). So two easy paths:

- **Drop it** (`gpio check --fix`) — the guide's default recommendation for 2.0, since native row-group statistics already give you the spatial filtering, and you'd save ~8% on every file; or
- **Declare it** (`gpio add bbox-metadata`) — keeps compatibility with readers that predate the native geospatial statistics.

Dropping is probably right here given how good the native statistics already look.

### The gap: STAC has no data assets

The guide recommends describing the data with STAC "using `application/vnd.apache.parquet` as the media type". That media type doesn't appear anywhere in the catalog — the only asset on any collection is `thumbnail` (image/png). The data is reachable via the `partition:glob` field and an `alternate` link to the HTML listing, but a generic STAC client can find the collections and not the files. The per-region `_manifest.json` files do carry per-theme bboxes, so the information exists; it's just not exposed as STAC assets or Items.

---

(end ai stuff)
For the STAC, I'm realizing that you implemented Portolan and we have a pretty big hole, in that we don't require to actually link to the parquet data. I'm going to work to add that to the spec / rashid. Thanks for your early testing of portolan - we definitely should follow the recommendation to always link to the actual parquet.

I updated the line in the commit to more accurately reflect things. Once you add the covering we can tweak what it says. Thanks for doing this! Will be good to have a good exemplar one for GeoParquet 2.0

gsueur:
Thanks for the review, both gaps were real and both are now fixed in the pipeline (merged today, so the published files and catalog carry them from tomorrow's nightly onward).

## The covering is declared now

You read it right: the bbox column predates the covering becoming optional, and I kept it to show what the explicit bbox buys a reader that does not understand native geo statistics. DuckDB's writer does not emit a `covering` (the field is not in GeoParquet 2.0 yet), so the pipeline now writes the whole `geo` value itself through `KV_METADATA`. Measured on the Rhode Island roads file, before and after:

| | before | after |
|---|---|---|
| size | 11.73 MB | 11.73 MB |
| row groups | 3 | 3 |
| bloom filters | 18 columns | 18 columns |
| native GEOMETRY | yes | yes |
| `geo` keys | one | one |
| row order | identical (compared the first 2,000 ids) | |

One note on `gpio add bbox-metadata`, since your review suggests it: its help says it "modifies the file in-place, preserving all data and file properties", but it rewrote that file from 11.73 MB to 15.93 MB, +36%, because it re-encodes at its own compression level rather than the file's (we write zstd 15). Worth a look, or at least a wording change in the help text.

## Confirming the page index

No `ColumnIndex` or `OffsetIndex` on any column, exactly as you said, and a control re-write with DuckDB reproduces it. So here the covering buys reader compatibility rather than page-level pruning, and the guide's wording should probably say that for a DuckDB-written file.

It did have one effect I did not expect: rashid reads spatial ordering *from* the covering column, so without one it reported "spatial ordering could not be evaluated". With the covering declared, `PTL-DAT-006` and `PTL-DAT-007` both disappear and the ordering check runs and passes.

## STAC data assets

Agreed, and that is a fair hit. Each collection now carries one asset per region:

```json
"data-us-ri": {
  "href": "https://parquetry.geomermaids.com/latest/country=US/state=US-RI/roads.parquet",
  "type": "application/vnd.apache.parquet",
  "title": "Rhode Island (US-RI)",
  "roles": ["data"],
  "alternate": {"s3": {"href": "s3://parquetry/latest/country=US/state=US-RI/roads.parquet"}},
  "file:size": 11227522
}
```

`partition:glob` stays as the bulk path. I did not model partitions as Items, since Portolan says not to for hundreds of partitions and this is 98 regions per theme, but that is the part your spec change may want to pin down.

One wrinkle worth feeding into the spec work: `file:checksum` is fine on a pinned, immutable snapshot, but the live catalog's hrefs point at a `latest/` alias that the next nightly replaces, so a checksum there would be stale by definition. We publish checksums only on the dated catalogs and omit them on the live one, following the "a publisher that cannot keep one current should omit it" line in Portolan's Assets section. Pinned catalogs validate with no warnings; the live one keeps `PTL-AST-003` warnings by design.

Happy to test whatever you land in the spec and rashid, this catalog is rebuilt nightly so it is cheap for us to move.

PR #299, opened 2026-08-18
Add LOD extension proposal

mmgeorge:
https://github.com/opengeospatial/geoparquet/issues/291

Split from https://github.com/opengeospatial/geoparquet/pull/295. 

This PR adds support for a new LOD extension. The geometry column definition is updated to support a new `lods` property. When defined, the array defines a set of other geometry columns that include simplified geometries and their associated resolution. Clients can use these reduced resolution geometries to avoid pulling in more geometric detail than is required for a given scale.


format-specs/lod.md
```diff
@@ -0,0 +1,112 @@
...
+    "lod_0": {
+      "path": ["lods", "level_0"],
+      "encoding": "WKB",
+      "geometry_types": ["Polygon", "MultiPolygon"]
+    },
+    "lod_1": {
+      "path": ["lods", "level_1"],
+      "encoding": "WKB",
+      "geometry_types": ["Polygon", "MultiPolygon"]
+    }
```
paleolimbot, line 109: I'm not sure I mind in either direction, but another option is to inline the LOD information into `geometry->lods[i]` directly (i.e., `{..., "lods": [{"path": [], "encoding": []`). They could be declared as top-level geometry columns, too, but we don't currently have a way to support the "path" for top level geometry columns.

paleolimbot (commented):
I've already given a bit of feedback so I'll let others take a look. I like this approach because it keeps attributes aligned in Parquet land (i.e., fewer non-Parquet tools required to render or perform a computation at a lower level of detail).

m-mohr:
How is this related to the proposal presented here at https://github.com/Kanahiro/cloud-optimized-geoparquet ?
If they are similar, I feel like we need to learn more about the approaches and then consider deciding for one depending on advantages and disadvantages. I don't really want two flavors of the same kind in the end...

PR #301, opened 2026-08-20
Remove stale 2D+M bbox text from schema.json

### LINKED ISSUE #300: Potential mismatch between md and schema re XYZ/XYM
Per https://github.com/opengeospatial/geoparquet/blob/v2.0.0-rc.1/format-specs/geoparquet.md?plain=1#L190-L194, it looks like a 6-value bbox always means XYZ, and cannot be XYM.

Per https://github.com/opengeospatial/geoparquet/blob/v2.0.0-rc.1/format-specs/schema.json#L66, XYZ or XYM are both acceptable.

My assumption is the text prevails and the "or 2D+M bbox (xmin, ymin, mmin, xmax, ymax, mmax)" should just be removed. This looks like it was settled in #278: the draft originally allowed a 6-value XYM bbox in both the text (a413f4c0e6474e8a1e366e1733f9b39db2e0c40c) and the schema (bcb49d6d675162fafff5251c4e0c4c97ba556773), then [this review suggestion](https://github.com/opengeospatial/geoparquet/pull/278#discussion_r2738295785) removed the XYM form from the text (applied in c83f2e9681) to keep the 6-value form unambiguous, but the schema description never got the matching removal there or in #287.

willcohen:
PR #278 removed the 6-element XYM bbox from the prose (c83f2e9), but the schema.json description of the 6-element form still says "or 2D+M bbox". This aligns the schema description with the prose: 6 elements always mean XYZ.

Fixes #300.

### DIFF
--- a/format-specs/schema.json
+++ b/format-specs/schema.json
@@ -63,7 +63,7 @@
                   "maxItems": 4
                 },
                 {
-                  "description": "3D bbox consisting of (xmin, ymin, zmin, xmax, ymax, zmax) or 2D+M bbox (xmin, ymin, mmin, xmax, ymax, mmax)",
+                  "description": "3D bbox consisting of (xmin, ymin, zmin, xmax, ymax, zmax)",
                   "minItems": 6,
                   "maxItems": 6
                 },


cholmes (approved):
Looks great, thanks for the contribution!

willcohen:
Thank you!

PR #298, opened 2026-08-18
Add PBF encoding extension proposal

mmgeorge:
**TODO: still need to add a script to generate data for this.**

https://github.com/opengeospatial/geoparquet/issues/291

Split from https://github.com/opengeospatial/geoparquet/pull/295. 

This PR adds support for a new PBF encoding. The intended use for this encoding is to support a new efficient encoding for web client display. Compared to WKB, PBF can be more than 3x smaller for a given LOD. 

I've extracted just the comparisons to WKB below, but some additional comparisons to other possible geometry formats can be found in https://github.com/opengeospatial/geoparquet/pull/295#issuecomment-5184148585. A deeper comparison to native quantized (the second best encoding) can be found in https://github.com/opengeospatial/geoparquet/pull/295#issuecomment-5287434585.

## Compressed Size by LOD, snappy

| Format | Encoding | Full (MiB) | LOD total (MiB) | level_2 | level_4 | level_6 | level_8 | level_10 | level_12 | level_14 | level_16 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| wkb | delta_byte_array | 218.48 | 291.85 (+0%) | 0.38 (+0%) | 1.12 (+0%) | 3.43 (+0%) | 9.11 (+0%) | 23.87 (+0%) | 52.50 (+0%) | 88.08 (+0%) | 113.23 (+0%) |
| pbf | delta_byte_array | 218.48 | 76.21 (-74%) | 0.14 (-63%) | 0.35 (-69%) | 0.94 (-73%) | 2.39 (-74%) | 5.75 (-76%) | 12.65 (-76%) | 22.80 (-74%) | 31.12 (-73%) |

## Compressed Size by LOD, gzip

| Format | Encoding | Full (MiB) | LOD total (MiB) | level_2 | level_4 | level_6 | level_8 | level_10 | level_12 | level_14 | level_16 |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| wkb | delta_byte_array | 157.94 | 177.52 (+0%) | 0.20 (+0%) | 0.62 (+0%) | 1.94 (+0%) | 5.12 (+0%) | 14.12 (+0%) | 31.84 (+0%) | 54.22 (+0%) | 69.40 (+0%) |
[...]

PR #302, opened 2026-08-22
Bring back the optional bbox covering from GeoParquet 1.1

cholmes:
## What this does

Restores the `covering` construct that GeoParquet 1.1 defined and the initial 2.0 spec dropped, as an **optional** column metadata field.

- **`format-specs/geoparquet.md`**: restores the `covering` table row, `#### covering` section, `##### bbox covering encoding`, and `### Bounding Box Columns` — all using the 1.1 spec language **verbatim** (verified by diff against `v1.1.0`; the only formatting change is a `json` tag on the code fence, matching current repo convention). A new note (the only new spec text) briefly explains the relationship to the native Parquet geospatial statistics.
- **`format-specs/schema.json`**: restores the 1.1 `covering` definition verbatim.
- **`scripts/test_json_schema.py`**: restores the 1.1 covering valid/invalid test cases verbatim (70 tests pass).
- **`format-specs/distributing-geoparquet.md`**: adds a "When to add a bbox covering column" section, and updates the Further Discussion page-level statistics writeup to point at it.

## Why

The native Parquet `GEOMETRY`/`GEOGRAPHY` geospatial statistics provide much of the original motivation for the bbox covering — readers can skip non-intersecting row groups with no extra column. But those statistics exist only at the **row group** level. A bbox covering is an ordinary struct column, so it gets a standard Parquet page index whose per-page min/max values let readers prune individual pages *within* a row group — early benchmarks in #279 roughly halved a selective query (~93 ms → ~48 ms) on a ~10M row file. It also keeps efficient spatial filtering working in readers that predate the native statistics. Making the covering optional in 2.0 lets publishers opt into those benefits without reverting to 1.1.

Since this adds an optional field that can be safely ignored, it is a forward-compatible change per the Version Compatibility rules (a minor-release change; the version number is left untouched here for the maintainers to bump at release time).

https://claude.ai/code/session_01FeF4uXsfjAVppmxy3B5ryB


format-specs/geoparquet.md
```diff
@@ -195,6 +196,49 @@ It is not currently possible to specify M bounds without Z bounds using a GeoPar
...
+
+The format of the `bbox` encoding is `{"xmin": ["column_name", "xmin"], "ymin": ["column_name", "ymin"], "xmax": ["column_name", "xmax"], "ymax": ["column_name", "ymax"]}`. The arrays represent Parquet schema paths for nested groups. In this example, `column_name` is a Parquet group with fields `xmin`, `ymin`, `xmax`, `ymax`. The value in `column_name` MUST exist in the Parquet file and meet the criteria in the [Bounding Box Column](#bounding-box-columns) definition. In order to constrain this value to a single bounding group field, the second item in each element MUST be `xmin`, `ymin`, etc. All values MUST use the same column name.
+
+The value specified in this field should not be confused with the top-level [`bbox`](#bbox) field which contains the single bounding box of this geometry over the whole GeoParquet file.
+
+Note: This technique to use the bounding box to improve spatial queries does not apply to geometries that cross the antimeridian. Such geometries are unsupported by this method.
```
paleolimbot, line 234: I know this is the exact text in the previous spec, but it's not clear what a reader should do if this happens (only really applies to GEOGRAPHY / spherical edges). Two options might be to suggest that implementations should error or that implementations should write -180..180 for the xmin..xmax bounds.

paleolimbot (approved):
Thank you!

yutannihilation:
> Making the covering optional in 2.0 lets publishers opt into those benefits without reverting to 1.1.

Maybe a naive question. Now that this PR is reviving covering, and [quite a few people seem to disagree with the idea that WKB should be the only GEOMETRY representation](https://github.com/opengeospatial/geoparquet/pull/298), is there much incentive to move to 2.0?

(Sorry if this came across as harsh. I'm genuinely just trying to understand what 2.0 is aiming for.)

yutannihilation:
Thanks! 

> I am not sure there's consensus that we should move away from WKB for the vast majority of full resolution geometry.

Ah, then it seems I simply misunderstood about this point. Sorry.

> the file sizes are smaller and more tools (notably: Spark) will be able to read/prune without extra bbox queries.

I think this explains the benefit of getting rid of the `covering` column. Yes, I agree with this point. 

I meant, those who want to keep relying on `covering` can keep using v1.1. I'm wondering how it would be different between putting `covering` into v2 and sticking with v1.1. (I vaguely assume GeoParquet v1.1 spec doesn't prohibit to use Parquet's native `GEOMETRY` type. If so, v1.1 can provide both `covering` and the row-groups-level statistics).

> I am hoping we can solve them in a way that can be upstreamed into the Parquet spec.

I really hope so!

cholmes:
> I meant, those who want to keep relying on covering can keep using v1.1. I'm wondering how it would be different between putting covering into v2 and sticking with v1.1. (I vaguely assume GeoParquet v1.1 spec doesn't prohibit to use Parquet's native GEOMETRY type. If so, v1.1 can provide both covering and the row-groups-level statistics).

Unfortunately 1.1 readers can't handle parquet's native geometry type. If there had been a way to keep things compatible, for older readers to understand the wkb in the native geometry type but rely on the geo metadata, then we likely would have just done a version 1.2 to explain that. We're doing 2.0 since it is a real change, but aiming to keep everything else the same. 

The main goal for 2.0 really is to align with the Parquet native geometry / geography types, and there was just no way to keep it compatible without the breaking change. It should be a pretty easy change for any library, but files will need to cut over to 2.0 if they want to take advantage of the new stuff. We also aim to encourage people to just use the native geometry / geography types  - you just need the geoparquet metadata if you want 'more' in a few small directions.

format-specs/geoparquet.md
cholmes: Yeah, I just copied the text exactly, but I'm fine for us to refine / improve further. I think suggesting writing -180 / 180 could be good - I agree the current text leaves you without a way forward, and it'd be nice to suggest something.

jorisvandenbossche:
> Unfortunately 1.1 readers can't handle parquet's native geometry type. If there had been a way to keep things compatible, for older readers to understand the wkb in the native geometry type but rely on the geo metadata,

FWIW, I think most readers can handle that fine nowadays (e.g. Arrow C++ / pyarrow was updated to ignore unknown logical types (and thus in this case would fall back on reading the physical bytes of the WKB column) before actual support for the geometry logical type was added; and I recall other readers that errorred on unknown logical types have been updated similarly). 

(the older versions of those readers of course still have that problem, and so I think that still warrants calling this 2.0)

format-specs/distributing-geoparquet.md
```diff
@@ -65,6 +65,30 @@ see [Further Discussion: page-level spatial statistics](#page-level-spatial-stat
 > wider range of software/versions to read your data. Many tools (GDAL, DuckDB, Hyparquet, geoparquet-io, SedonaDB, QGIS)
 > will work with Parquet native types (and the GeoParquet 2.0 metadata), and eventually 2.0 will be the only recommended way.
 
+### When to add a bbox covering column
+
+GeoParquet 2.0 retains the [`bbox` covering](geoparquet.md#covering) from version 1.1 as an option. Most files don't
+need it — the native row group statistics described above provide efficient spatial access without any extra column — so
+the default recommendation is to leave it out. But there are two situations where adding it can be worth the additional
+file size:
```
jorisvandenbossche:
```suggestion
need it because the native row group statistics described above provide efficient spatial access without any extra column.
Therefore, the default recommendation is to leave it out, but there are two situations where adding it can be worth the additional
file size:
```

format-specs/distributing-geoparquet.md
```diff
@@ -65,6 +65,30 @@ see [Further Discussion: page-level spatial statistics](#page-level-spatial-stat
...
+
+- **Page-level spatial pruning.** The native geospatial statistics only exist at the row group level, so once a row
+  group matches a query the reader must fetch all of it. A `bbox` covering is an ordinary Parquet struct column, so it
+  gets a standard [page index](https://github.com/apache/parquet-format/blob/master/PageIndex.md) whose per-page min/max
+  values let readers skip individual pages *within* a row group. Early benchmarks
+  ([issue #279](https://github.com/opengeospatial/geoparquet/issues/279)) saw a selective spatial query on a ~10 million
+  row file roughly halve in time (~93 ms down to ~48 ms) with page-level pruning. The benefit is largest for highly
```
jorisvandenbossche: I personally would not include this link. Such concrete numbers can go outdated very quickly, and also just depend a lot on the exact details

jorisvandenbossche (approved):

format-specs/distributing-geoparquet.md
jorisvandenbossche: Also, those numbers might actually be hallucinated? I don't see them in that issue. Or you meant to link elsewhere?

(the arguments in the issue are also not really about page-level statistics, while here it is mentioned in that bullet point)

yutannihilation:
> FWIW, I think most readers can handle that fine nowadays

Thanks so much for clarifying. This answers what I was wondering. I was expecting this could be absorbed on the implementation's side without requiring the specification changes. But I now agree that this is 2.0 in the sense that this is still a breaking change for the older readers.

Sorry for the noise!

format-specs/distributing-geoparquet.md
cholmes: I'll just remove that sentence. Thanks for the close read.

jatorre:
Running the OGC abstract tests on real files (geoparquet/geoparquet-testing#6) turned up a mismatch between this text and the largest producer. Both 1.1.0 and this PR say of the bounding box column: "the child fields MUST be named `xmin`, `ymin`, `xmax`, and `ymax` and MUST be ordered in this same way." Every Overture Maps release writes them as `xmin, xmax, ymin, ymax` (checked on 2026-08-19.0, all themes). No validator has ever checked the order, so nobody noticed; GDAL, GeoPandas and SedonaDB follow the text.

Parquet readers address struct fields by name, so the order carries no information for a reader. I'd suggest dropping "and MUST be ordered in this same way" here, keeping the names as the requirement. The alternative is asking Overture to rewrite, which seems disproportionate for a rule with no functional effect. Recorded as SI-26 in #304.

The idea of relying in order sounds fragile

jatorre:
One more data point on the field order: the 1.1.0 specification's own `examples/example.parquet` writes the bbox struct as `xmax, xmin, ymax, ymin`, and Apache Sedona's 1.1 test file does the same. So the "ordered in this same way" sentence was not followed even by the reference example, which I think settles it: keep the names, drop the order.

cholmes:
Yeah, sounds good to just drop, especially since 'Parquet readers address struct fields by name'.

format-specs/geoparquet.md
```diff
@@ -195,6 +196,49 @@ It is not currently possible to specify M bounds without Z bounds using a GeoPar
...
+
+### Bounding Box Columns
+
+A bounding box column MUST be a Parquet group field with 4 or 6 child fields representing the geometry's coordinate range. For two-dimensional data, the child fields MUST be named `xmin`, `ymin`, `xmax`, and `ymax` and MUST be ordered in this same way. As with the top-level [`bbox`](#bbox) column, the values follow the GeoJSON specification (RFC 7946, section 5), which also describes how to represent the bbox for geometries that cross the antimeridian. For three dimensions the additional fields `zmin` and `zmax` MAY be present but are not required. If `zmin` is present then `zmax` MUST be present and vice versa. If `zmin` and `zmax` are present, the ordering of the child fields MUST be `xmin`, `ymin`, `zmin`, `xmax`, `ymax`, `zmax`. The fields MUST be of Parquet type `FLOAT` or `DOUBLE` and all columns MUST use the same type. The repetition of a bounding box column MUST match the geometry column's [repetition](#repetition). A row MUST contain a bounding box value if and only if the row contains a geometry value. In cases where the geometry is optional and a row does not contain a geometry value, the row MUST NOT contain a bounding box value.
```
cholmes: Just drop it and not mention order?

```suggestion
A bounding box column MUST be a Parquet group field with 4 or 6 child fields representing the geometry's coordinate range. For two-dimensional data, the child fields MUST be named `xmin`, `ymin`, `xmax`, and `ymax`. As with the top-level [`bbox`](#bbox) column, the values follow the GeoJSON specification (RFC 7946, section 5), which also describes how to represent the bbox for geometries that cross the antimeridian. For three dimensions the additional fields `zmin` and `zmax` MAY be present but are not required. If `zmin` is present then `zmax` MUST be present and vice versa. If `zmin` and `zmax` are present, the ordering of the child fields MUST be `xmin`, `ymin`, `zmin`, `xmax`, `ymax`, `zmax`. The fields MUST be of Parquet type `FLOAT` or `DOUBLE` and all columns MUST use the same type. The repetition of a bounding box column MUST match the geometry column's [repetition](#repetition). A row MUST contain a bounding box value if and only if the row contains a geometry value. In cases where the geometry is optional and a row does not contain a geometry value, the row MUST NOT contain a bounding box value.
```

Or does anyone want us to explicitly say that they can be in any order?

jatorre:
Id rather not mention order... soimpler is better

Issue #277, opened 2025-09-24
Broken link in README (OGC candidate Standard)


severo:
The following link is broken:

> The OGC candidate Standard is at https://docs.ogc.org/DRAFTS/24-013.html.

m-mohr:
The link just doesn't exist anymore. The annex part is something different. I think this part should just be removed for now.

PR #224, opened 2024-05-29
Migration of the GeoParquet specification to the official OGC templates

m-mohr:
Based on #219 and #206. Original author is @ghobona.
Making this a draft for now because the CI is failing, but I don't have the time and knowledge about metanorma to fix this.

Adds back the following commits:
- 5d67a9c5
- 471a2f88
- 2aa06683
- a727140f

The docs may need to be updated with the commits that were merged afterwards:
https://github.com/opengeospatial/geoparquet/compare/fe3687dfa732b47e7891b0e235febf6c10fc03f1...main

PR #304, opened 2026-09-03
OGC Standard document for GeoParquet 2.0 (metanorma), aligned with the 2.0.0 release

jatorre:
## Why

The OGC GeoParquet SWG needs an OGC-formatted version of the specification (metanorma, OGC
Standard template) to take to the OGC Architecture Board and to a membership vote. That work
stalled in December 2024 with #224 (@m-mohr, based on @ghobona's migration) and #247 (@cholmes),
both written for GeoParquet 1.0. With 2.0.0 about to be released, the SWG (with OGC staff)
agreed to target **2.0** directly: 1.x got the community to implement, 2.x is what implementers
should adopt, and it is far less likely to need another major revision. OGC membership will be
notified separately with a short note explaining this.

**This PR is not on the 2.0.0 release path and needs no action before the tag.** It changes
nothing under `format-specs/`.

## What is in the PR

Everything lives in a new `ogc/` directory plus `docs/ogc/`:

* `ogc/` — the metanorma document for GeoParquet **2.0.0**. Two requirements classes, both
  targeting the file:
  * **Core**: 20 requirements, 6 recommendations, 3 permissions;
  * **Bounding Box Covering**: 6 requirements, taken from the #302 text and marked conditional
    on that PR merging.
  * One abstract test per requirement (26). Every requirement file starts with a `// Source:`
    comment pointing at the line of `geoparquet.md` it restates.
  * Normative references pin Apache Parquet 2.12.0 (`LogicalTypes.md`, `Geospatial.md`) instead
    of restating the encoding.
  * The community spec's expectations on *readers* (interpret PROJJSON and authority:code
    `crs`, don't reject unknown fields, forward compatibility) are informative (Clause 8). The
    SWG chose a single standardization target, the file, to keep approval simple.
* `docs/ogc/CHANGES.md` — **start here if you review.** Every place where the OGC wording
  differs from `geoparquet.md`, with the reason. The intent is that the OGC text is purely
  derivative; where the markdown implies something without stating it, the OGC text has a NOTE,
  not a requirement.
* `docs/ogc/spec-issues.md` — things in the community spec that the drafting surfaced. Two are
  one-line edits worth making before the 2.0.0 tag (a Parquet link that points at a version
  without the text it cites; "spherical" vs "not planar" for the orientation recommendation).
  The rest are for after the release. Nothing was changed in `format-specs/`; I'll open issues
  for whichever of these the maintainers think are real.
* `docs/ogc/gaps.md` — for each abstract test, which existing validator (GPQ, GDAL
  `validate_geoparquet.py`, `geoparquet-testing`) covers it. Eight of the 26 tests have no
  implementation anywhere (four of them in the covering class); nine are fully covered.
* `docs/ogc/AUDIT.md`, `docs/ogc/crosswalk.md` — the audit of #224/#247 against 1.1 and 2.0
  that this work started from.
* `.github/workflows/ogc-document.yml` — builds the document with the `metanorma/metanorma`
  Docker image on pushes/PRs touching `ogc/**`; fails on metanorma errors or a requirement
  without a test; uploads the HTML. The old `docs.ogc.org/DRAFTS/24-013` cron build is gone
  (#277), so this replaces it.

The #224 and #247 branches are merged into this branch so the original authorship is preserved
in history; the files were then moved from `format-specs/` to `ogc/` and rewritten for 2.0.

## Questions for reviewers

1. **Editors and contributors.** `document.adoc` lists Chris Holmes, Tim Schaub, Joris Van den
   Bossche, Kyle Barron and Javier de la Torre as editors (carried over from #224). Who else
   should be there, and who should be in the Contributors table? @cholmes, your view first.
2. **Location.** `ogc/` rather than `format-specs/` (where #224 had it), so the community spec
   directory stays as it is and CI can be scoped. Happy to move it back.
3. **#302.** If the covering does not come back, Clause 7 and its conformance class are removed
   as a unit.
4. **Document number 24-013** was assigned for the 1.0 draft; I'll confirm with OGC staff that
   it carries over.
5. **CI.** The workflow pulls the metanorma image (~2 GB) on each run touching `ogc/`. Fine, or
   should it be manual-only?

## How to build locally

```sh
cd ogc
docker run --rm -v "$PWD":/metanorma metanorma/metanorma \
  metanorma compile --agree-to-terms -t ogc -x xml,html document.adoc
open document.html
```

Supersedes #224 and #247. Refs #277, #302.


jatorre:
@cholmes thinking more about this.... I rally thing that Parquet has these options for a reason.

I would only keep geospatial statistics present, and rows spatially ordered as mandatory, and leave the 150K row group and bbox as recommendations....

I just feel we are not gonna find the best sweetspot and likely will change over time... and the bbox too i can see how it feel intrusive to me.

Are we REALLY sure we want to be this prescriptive?? I feel is gonna be a pain to have engines output "correct" geoparquet, so we are gonna end up with not fully supported geoaprquet but kind of geoparqeutish... 

-Bbbox is adding extra columns, makes it really hard for imports and export engines as we enfgorce some form of schema
-row group there could be cases where that number does not make sense...
-spatiallñy ordered... what if the geoaprquet is being created in batches... the ordered would be lost... so it would be "geoparquet" compliant sometimes, not always...

I think we are putting ourselves in too much of a trouble....

cholmes:
> I would only keep geospatial statistics present, and rows spatially ordered as mandatory, and leave the 150K row group and bbox as recommendations....

> I just feel we are not gonna find the best sweetspot and likely will change over time... and the bbox too i can see how it feel intrusive to me.

> Are we REALLY sure we want to be this prescriptive?? I feel is gonna be a pain to have engines output "correct" geoparquet, so we are gonna end up with not fully supported geoaprquet but kind of geoparqeutish...

So to me the idea is that it's just a 'profile', a totally optional thing. The 'core' is 'correct geoparquet' and says nothing about row group size, spatial ordering, compression, or bbox. A profile just lets people say that they've followed a set of rules. It's a pretty common pattern in OGC, just a group of extra requirements.

But honestly I don't really care what happens in this ogc spec area (just want it to be consistent and not embarrassing) , so if you want to not do the 'profile' thing that's fine with me - just leave row group size and spatial ordering out entirely (and bbox is just an optional construct in the core, but with no recommendation to use it).

jatorre:
Agreed, let's leave it as it is: the Cloud-Optimized Distribution class stays an optional profile with two requirements (geospatial statistics present, rows spatially ordered) and everything else (row group size, compression, bbox covering, partitioning, STAC) as recommendations. Core says nothing about any of it. That is the middle ground I was asking for, and the validator already reports the class that way, with the recommendations as advice rather than verdicts.

m-mohr:
I'm actually inclined go stricter. The standard should only include the actual standard parts and the profiles, not the (current) best practices.
We can create a separate OGC best practice document if needed, otherwise I feel we are not certain enough about the best practices yet (especially row group size, partitioning, compression). So baking them into the standard may bite us later, I feel.
Best practices can be updated better than standards...

jatorre:
Ok. Im convinced too. The 2 requirements of the profile (stats + spatia order) are the part we are sure about, and the 6 recommendation (row group, compression, bbox,partitioning, STAC) can be advice as it might change over time. So we move it to a community guide. 

All in favor?

ogc/sections/clause_8_distribution.adoc
```diff
@@ -0,0 +1,75 @@
...
+
+=== Spatial ordering
+
+Row-group statistics only help when the rows of the file are spatially ordered, so that each row group covers a compact part of the extent and a query touches few of them. GeoParquet does not prescribe an ordering method: Hilbert curve, GeoHash, S2 or R-tree based orderings all work. The requirement is stated on the outcome, the pruning that the row-group bounding boxes allow, not on a particular ordering.
+
+include::../requirements/distribution/req_spatial-order.adoc[]
+
+NOTE: How much pruning counts as "most" depends on the number of row groups: two row groups can never let a reader skip more than about half of the file, while several hundred well-ordered row groups let it skip nearly all of them. The executable test therefore compares the pruning a file achieves with the pruning an ideal tiling into the same number of row groups would achieve, and the threshold is maintained in the test suite rather than in this Standard, because the community is still refining it (Portolan specification and `geoparquet-io`, 2026).
+
+=== Row group size
```
jorisvandenbossche: This and the ones below should be removed now?

PR #305, opened 2026-09-19
Propose a Level of Detail (LoD) extension for progressive feature access

Kanahiro:
This PR proposes optional `geo.lod` metadata and a coarse-to-fine row group layout, allowing readers to fetch features progressively based on rendering resolution.
Each row is stored once, preserving geometries and attributes. Readers can ignore the extension and read ordinary GeoParquet.

This proposal is related to #299, which proposes geometry-level LoD. However, geometry-level LoD alone does not reduce the number of features to fetch. This PR therefore starts with feature-level LoD: progressively adding features without changing their geometries. I plan to extend this to geometry-level LoD in the future.

Reference implementation: https://github.com/Kanahiro/cloud-optimized-geoparquet/tree/3541e4f9bcd4fe623d0fb16b4b7ac462ca2937b0


Kanahiro:
@paleolimbot 
Thank you for reading my proposal.

> As you noted, this is orthogonal to the initial LOD proposal...

Your understanding is completely correct. 
Besides, I think geometry-level LoD should be based on feature-level LoD mecanism because visually "simplifying" doesn't only mean reducing vertices but dropping/skipping features which have no visual importance, as MVT does. Then, I think feature-level/geometry-level LoD are not orthogonal and geometry-level LoD should depend on feature-level LoD structure.

> Instead of putting this in metadata...

Good point! Actually in my first poc, I introduced "zoomlevel" column indicating resolutions of each row.

https://github.com/Kanahiro/yosegi

Drafting and sophitsicating this spec, I realized that information of "resolution" can be expressed only in metadata without modifying original table structure. I feel this more generic and clean because such column has quite low cardinality and adding it to entire table seems not to be good to me. A reader can recognize feature-level LoD structure only with metadata.
However, it should be taken into account that writers can not always easily modify "geo" metadata as you pointed.

PR #306, opened 2026-09-21
Update compatible Parquet guide for GeoParquet 2.0

cholmes:
## What this changes

`format-specs/compatible-parquet.md` was written for GeoParquet 1.1 and still carried a "not updated for 2.0" warning. This PR rewrites it around the GeoParquet 2.0 story:

- **New framing up front**: any Parquet file that uses the native Parquet `GEOMETRY`/`GEOGRAPHY` logical types is completely compatible with the GeoParquet ecosystem — every GeoParquet 2.0 reader must read it. It is not generally necessary to fully implement GeoParquet 2.0, and a new section explains the situations where adding the full `geo` metadata is worth it (guaranteed inline PROJJSON CRS, winding order and epoch, multiple geometry columns, bbox covering, general-purpose tools).
- **A new "When you need the compatibility guidelines" section** that distinguishes three cases: tools that write the native types (done), geospatial tools that write GeoParquet 1.x (also valid GeoParquet, no compat guidelines needed), and general-purpose Parquet tools with no geospatial support — the case the guidelines are actually for.
- **The original recommendations are kept**, moved into a clearly scoped "Compatibility guidelines for tools that pre-date the Parquet geospatial types" section, with the reader-fallback advice updated to check native types first.

An earlier revision included a table of the library versions that first supported the native geospatial types; it was dropped to avoid a maintenance burden as support evolves.


format-specs/compatible-parquet.md
```diff
@@ -1,18 +1,97 @@
...
+
+* An unambiguous, self-describing declaration that the column contains geospatial data (no guessing
+  from column names or sniffing binary blobs).
+* The coordinate reference system carried on the column itself, via the logical type's `crs` property.
+* [Geospatial statistics](https://github.com/apache/parquet-format/blob/apache-parquet-format-2.12.0/Geospatial.md#geospatial-statistics)
+  — per-row-group bounding boxes and geometry type lists — that let readers efficiently skip data
+  that does not match a spatial query.
+
+If you are producing data this way, GeoParquet readers will read it. The one recommendation that goes
+beyond what the core Parquet specification requires: if your tool lets you control the `crs` property,
```
paleolimbot:
```suggestion
beyond what the core Parquet specification requires: if your tool lets you control the `crs` property of the Parquet logical type,
```

format-specs/compatible-parquet.md
```diff
@@ -1,18 +1,97 @@
...
+* **The file has multiple geometry columns.** The `primary_column` field tells readers which one to
+  use by default. This isn't necessary for many use cases, but it can help renderers and GIS tools know
+  which column to use first.
+* **You want a [bbox covering](./geoparquet.md#covering) column.** The covering metadata points
+  readers at a per-row bounding box column, which enables page-level spatial pruning and spatial
+  filtering in readers that pre-date the native geospatial statistics — see
+  [when to add a bbox covering column](./distributing-geoparquet.md#when-to-add-a-bbox-covering-column).
+* **You are writing a general-purpose geospatial tool or library.** Emitting the full metadata costs
+  little and maximizes what downstream readers can do with the data, so it is the recommended default
+  for tools whose job is producing geospatial output.
```
paleolimbot: Maybe:

```suggestion
* **You are writing a general-purpose geospatial tool or library.** Emitting the full metadata costs
  little and enables tools that do not support Parquet logical types (including many
  GeoParquet 1.0 and GeoParquet 1.1 readers) to reconstruct the geometry type
  so it is the recommended default for tools whose job is producing geospatial output.
```

(just trying to add that I think it's more about which readers can read rather than what they can do with it)

paleolimbot (approved):
A few optional suggestions...these make sense to me!

PR #307, opened 2026-09-21
Add changelog covering 1.0, 1.1 and 2.0

cholmes:
Adds a `CHANGELOG.md` at the repo root summarizing the three stable releases, focused on the meaningful changes an implementor needs rather than minor tweaks. The main goal is making the 1.1 → 2.0 differences easy to see, and the text is written to be usable directly in release notes.

- **2.0.0**: prose framing (native Parquet geospatial types as the foundation, `geo` metadata as the layer on top) followed by one-line PR-linked bullets: required `GEOMETRY`/`GEOGRAPHY` logical types, removal of the 1.1 GeoArrow encodings, new `edges` values, the CRS split between the Parquet `crs` property and inline-PROJJSON metadata, `srid:0`, readers accepting metadata-less native-type files, and the bbox covering staying optional (reflecting the post-rc restoration in #302).
- **1.1.0**: the bbox covering and GeoArrow encodings as the two major additions, plus the version compatibility policy, `crs: null` clarification, and file extension / media type recommendations.
- **1.0.0**: written as the start of the changelog — a summary of the spec the release established rather than a diff against the beta.


paleolimbot (approved):
I think this is a great summary...I am not sure if you want to include motivations for a few more changes (e.g., why arrow encodings were removed), but also this is great as just a summary and we can update this with clarification if required.

jorisvandenbossche:
Might want to reuse some parts from https://github.com/opengeospatial/geoparquet/releases/tag/v2.0.0-rc.1? (without the auto-generated "What's Changed" part) 

(because one of the goals here for me was also to have a reviewed text we can use for that announcement)

